Issue #10, opened 2020-04-04
[preset-env] Add ability to error if polyfills are required


davidje13:
## Feature Request

Some corejs polyfills are overly bloated (or have trivial alternatives in many situations) and it can be desirable to avoid the need for them. `preset-env` already has the ability to detect where polyfills would be required in code via `useBuiltIns` (and with more accuracy than a linter since it is the tool which actually adds the polyfill!)

Extending the existing concept of the `exclude` section to provide a way of forbidding individual polyfills would be useful for managing this:

```json5
{
  "useBuiltIns": "inline",
  "corejs": 3,
  "polyfills": {
    "web.url": "useBuiltIn", // or true
    "es.string.replace": "exclude",
    "web.immediate": "error"
  }
}
```

This would replace `exclude` (with a simple migration path) and has the advantage of having a less ambiguous name.

Unconfigured polyfills could default to either `useBuiltIn` (to match current behaviour and make it easier to adopt new polyfills) or `error` (to force developers to consider each newly required polyfill rather than blindly accepting it). I could see arguments for both defaults; not sure what's best (though I personally would prefer an `error` default).

ljharb:
I'm somewhat skeptical here; it's pretty impossible imo to confidently claim that nothing in your entire dependency graph cares about the edge cases, or exact spec compliance; I've seen that assumption bite far too many people in production.

nicolo-ribaudo:
I think this could be useful in two cases:
- `error` is useful if you know that the polyfill for X is slow/big and you want to be sure not to write any code that would need it
- if you are only polyfilling your own code (using `usage-global` or `usage-pure`, and not transpiling `node_modules`), then you can be confident that you don't need some edge cases to be handled.

However, I think that this (or at least, `error`) could be implemented by adding a lower-level function option that is called whenever a polyfill is injected. This also satisfies other use cases that have been requested in various channels:
- programmatically get a list of all the injected polyfills
- automatically `npm install` new polyfills (<-- This is something I personally don't like)

WebReflection:
Some polyfill is unreliable by specs, so I don’t see any guarantee one polyfill would grant any edge caae better than others, when specs don’t allow it (I.e. weak maps)

ljharb:
tbh I wouldn’t suggest including anything automatically that can’t be shimmed, like weak maps.

pleunv:
> 
> 
[...]

PR #13, opened 2020-05-18
Add a `shouldInjectPolyfill` option to polyfill providers

nicolo-ribaudo:
This new API has the same goal as the include/exclude options
(for this reason, it will throw an error if you try to use both), but
both have some advantages that the other one doesn't have:
1) include/exclude are an array, so you can use them
    in a static JSON file. Also, they might be easier to use.
2) shouldInjectPolyfill is a function, so it is more powerful: for
    example, you could use it to store somewhere a list of all
    the injected polyfill, or to throw an error if a specific
    polyfill would be injected (maybe because it's too big and
    you want to be sure that you don't rely on it)

Fixes https://github.com/babel/babel-polyfills/issues/10, fixes https://github.com/babel/babel/issues/11583 (cc @niftylettuce), fixes https://github.com/babel/babel/issues/6625,  can fix https://github.com/babel/babel/issues/8489

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -62,6 +62,18 @@ See the same options at https://babeljs.io/docs/en/babel-preset-env
 
 ## Advanced options
 
+### `shouldInjectPolyfill`
+
+`include` and `exclude` are great ways to forcefully mark a given polyfill as supported or unsupported.
+
+However, sometimes you might need something mroe advanced or customized: for example, you could want to store somewhere a list of the injected polyfills, or to throw an error when a given polyfill would be included.
+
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
For the people wondering about the issues closed by this PR in the babel/babel repository: we'll release this alongside with the next Babel release.

niftylettuce:
@nicolo-ribaudo can you ping me when this is released to npm?  do you mean next major, minor, or patch of `@babel/core`?

existentialism:
@niftylettuce you can always watch "releases only"

nicolo-ribaudo:
This has been released!

PR #15, opened 2020-05-25
Improve readme

nicolo-ribaudo:


README.md
```diff
@@ -1,15 +1,96 @@
 # Babel Polyfills
 
-A set of Babel plugins that enable injecting different polyfills with different strategies in your compiled code.
```
hzoo, line 3: add this back!

README.md
```diff
@@ -1,15 +1,96 @@
...
+> ⚠️ These packages are **highly experimental** and they have not been tested in production applications yet. Also, we are still working ond wiring some missing polyfills.
 
 > ℹ️ This repository implements what was initially proposed at [babel/babel#10008](https://github.com/babel/babel/issues/10008).
 
 > 💡 If you are looking for some quick setup examples, or just want to see how to migrade your config, please check [`docs/migration.md`](https://github.com/babel/babel-polyfills/blob/master/docs/migration.md).
 
-## How does it work?
+## Usage
+
+The main Babel packages only transform JavaScript _syntax_: you also need to load a polyfill, to make native _functions_ or _built-in objects_ work in older browsers. The Babel plugins implemented in this repository automatically inject the polyfills in your code, trying to only load what is really needed basd on your compilation targets and on what you are using in your code.
```
hzoo, line 14:
```suggestion
The main Babel packages only transform JavaScript _syntax_: you also need to load a polyfill, to make native _functions_ (`Array.prototype.flat`) or _built-in objects_ (`Promise`) work in older browsers. The Babel plugins implemented in this repository automatically inject these polyfills in your code, while trying to only load what is really needed. It does this based on your compilation targets and on what you are using in your code.
```

hzoo: Actually would like to add a short thing about how you would normally do `<script src=` or do `import "polyfill"` at the top of your file but that assumes your targets don't change or that you are using that polyfill.

hzoo: Could be good in history section too

README.md
```diff
@@ -3,13 +3,97 @@
...
+
+<!--prettier-ignore -->
+| Polyfill | Plugin | Methods |
+| :------: | :----: | :-----: |
+| `core-js@2` | [`babel-plugin-polyfill-corejs2`](./packages/babel-plugin-polyfill-corejs2) | `entry-global`, `usage-global` and `usage-pure` |
+| `core-js@3` | [`babel-plugin-polyfill-corejs3`](./packages/babel-plugin-polyfill-corejs3) | `entry-global`, `usage-global` and `usage-pure` |
+| `es-shims` | [`babel-plugin-polyfill-es-shims`](./packages/babel-plugin-polyfill-es-shims) | `usage-global` and `usage-pure` |
+| `regenerator-runtime` | [`babel-plugin-polyfill-regenerator`](./packages/babel-plugin-polyfill-regenerator) | `entry-global`, `usage-global` and `usage-pure` |
+
+> 💡 If you want to implement support for a custom polyfill, you can use `@babel/helper-define-polyfill-provider`. ([`docs/polyfill-provider.md`](https://github.com/babel/babel-polyfills/blob/master/docs/polyfill-provider.md).)
```
hzoo, line 101: You basically said this here, but above the table should we say, these are the ones we've made/maintained/setup but doesn't mean you have to use these? Or a note we can list others if it meets some criteria

README.md
```diff
@@ -3,13 +3,104 @@
...
+## Usage
+
+The main Babel packages only transform JavaScript _syntax_: you also need to load a polyfill, to make native _functions_ (`Array.prototype.flat`) or _built-in objects_ (`Promise`) work in older browsers.
+
+The easiest way to do so is to directly load the polyfill using a `<script ...>` tag:
+```html
+<script src="https://unpkg.com/core-js-bundle@3.6.5/minified.js"></script>
+```
+
+However, this simple approach can potentially include a lot of unnecessary code. The Babel plugins implemented in this repository automatically inject the polyfills in your code, while trying to only load what is really needed. It does this based on your compilation targets and on what you are using in your code.
```
hzoo, line 21: Maybe everything before `The Babel plugins implemented in this repository automatically ` could go in a `## Motivation/Reasoning/Why Section`?

nicolo-ribaudo: We have that section at the end, but I think that a few sentence i this Usage section (since it is what most people read) help giving a bit of context before reading everything without distracting too much with the long explanation.

README.md
```diff
@@ -3,13 +3,106 @@
...
+});
+ 
+```
+
+</td>
+<td>
+
+```js
+import "promise.prototype.finally/auto.js";
+import "promise.allsettled/auto.js";
```
ljharb, line 71:
```suggestion
import "promise.prototype.finally/auto";
import "promise.allsettled/auto";
```

Extensions should always be omitted in imports in general, but in this case the `exports` field will not allow the extensions in packages that support it.

nicolo-ribaudo: https://github.com/babel/babel-polyfills/commit/2fdfe1b2ba791279f7ef0e037599fe8951aa3704

nicolo-ribaudo:
Feel free to directly edit README on master if you see any other error.

PR #22, opened 2020-06-28
Move `String#replaceAll` to stage 4

pustovalov:
https://github.com/tc39/proposal-string-replaceall

Browser compatibility:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll


nicolo-ribaudo:
Thanks!

PR #23, opened 2020-06-28
[es-shims] Add compat data for `String.prototype.matchAll`

pustovalov:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll

### DIFF
--- a/scripts/build-es-shims-data/es-shims-features.js
+++ b/scripts/build-es-shims-data/es-shims-features.js
@@ -52,7 +52,7 @@ module.exports = {
     "String.prototype methods / String.prototype.endsWith",
   "String.prototype.includes":
     "String.prototype methods / String.prototype.includes",
-  "String.prototype.matchAll": [],
+  "String.prototype.matchAll": "String.prototype.matchAll",
   "String.prototype.padEnd": "String padding / String.prototype.padEnd",
   "String.prototype.padStart": "String padding / String.prototype.padStart",
   "String.prototype.repeat":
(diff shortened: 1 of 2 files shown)


nicolo-ribaudo:
Thanks for these PRs!

I'm preparing a PR to automatically generate compat data from `compat-table`: it's not finished yet, but I can upload it in ~30mins so we can finish it together if you want.

pustovalov:
I'll be glad to help

pustovalov:
hm, is this right path `String.prototype methods / String.prototype.matchAll` for `String.prototype.matchAll`?

scripts/build-es-shims-data/es-shims-features.js
```diff
@@ -52,7 +52,11 @@ module.exports = {
     "String.prototype methods / String.prototype.endsWith",
   "String.prototype.includes":
     "String.prototype methods / String.prototype.includes",
-  "String.prototype.matchAll": [],
+  "String.prototype.matchAll": {
+    features: [
+      'String.prototype.matchAll / basic functionality',
+    ],
+  },
```
nicolo-ribaudo, line 59: The `matchAll` polyfill from @es-shims [expliticly handles](https://github.com/es-shims/String.prototype.matchAll/blob/master/polyfill.js#L7-L9) the "throws on non-global regex" case so we should consider it.

Thus, since we would have to list every subcategory of `String.prototype.matchAll`, we can use the "shorthand" version:

```suggestion
  "String.prototype.matchAll": "String.prototype.matchAll",
```

pustovalov: changed

PR #26, opened 2020-07-06
Avoid generating empty identifiers, disallowed by Babel 8

nicolo-ribaudo:


.travis.yml
```diff
@@ -12,6 +12,8 @@ jobs:
   include:
     - name: "Test"
       before_script: yarn build
+      env:
+        BABEL_TYPES_8_BREAKING: true
```
ljharb, line 16: what does this do?

JLHwung: It is a feature flag which enables stricter AST type checking in Babel 8.

nicolo-ribaudo:
> What are "empty identifiers"?

`t.identifier("")`. It's something that doesn't make sense in JavaScript, and were created due to a logical error in the code (I was always creating the identifier even for `import "foo"`).

PR #31, opened 2020-08-21
fix: Promise.allSettled not included when it should be

jquense:
The other way to fix this would be to include it in shipped proposals. Effectively this is getting filtered out unless `proposals: true` is set when it shouldn't be since it's shipped and stage 4


jquense:
friendly ping :)

nicolo-ribaudo:
I'll release this weekend

jquense:
Thanks y'all !

nicolo-ribaudo:
Released

PR #39, opened 2020-10-14
Object.keys polyfill

pustovalov:


ljharb (approved):

nicolo-ribaudo:
The `object-keys` package does not conform to the [es-shims API](https://github.com/es-shims/es-shim-api) interface, so it won't work with the way this package injects polyfills (for example, `object-keys/auto` does not exist).

ljharb:
ohhh right, i forgot i haven't converted that one yet :-)

I can do that and release it as a breaking change this week.

ljharb:
(i still haven't done this; it's on my list)

PR #40, opened 2020-10-25
Implement `absoluteImports` option

nicolo-ribaudo:
This option is similar to `@babel/transform-runtime`'s `absoluteRuntime` option.
It has been renamed because polyfill providers don't injected imports to `@babel/runtime`.

PR for `@babel/transform-runtime`: https://github.com/babel/babel/pull/8435

This option is an improvement over `@babel/transform-runtime`'s, since it also works for polyfills attached to the global scope.

Ref: https://github.com/babel/babel/issues/12035


packages/babel-helper-define-polyfill-provider/src/index.js
```diff
@@ -97,6 +110,7 @@ function resolveOptions<Options>(
     method,
     methodName,
     targets,
+    absoluteImports: absoluteImports || false,
```
JLHwung, line 113: `""` will be cast into `false`. Is that desired?

JLHwung (approved):

arcanis:
Will this also benefit Babel 7, or only Babel 8+?

nicolo-ribaudo:
This plugin is already fully compatible with Babel 7: the only difference is that in Babel 8 it will be the only way to inject polyfillls.

We'll discuss if it makes sense to add an equivalent option, or if we'll only implement it here to encourage migration.

43081j:
Could you provide an example of how this might be configured? would we just set `absoluteImports: true` in preset-env's config? or is there an extra level needed

PR #41, opened 2020-11-09
Update `core-js-compat` and `compat-table`

nicolo-ribaudo:


zloirock (approved):

JLHwung (approved):
Test failure is due to out of dated test fixture.

PR #43, opened 2020-11-25
Ignore identifiers in "export ... from" declarations

### LINKED ISSUE #42: Unnecessary polyfilling + wrongly renaming imports
I babelify + auto polyfill node_modules too; because it's usually unclear what ES version+browser libraries target nowadays. After switching to `babel-polyfills` (corejs3), one curious case failed.

For esm5/esm2015 builds of `rxjs@6.6.3`; in `node_modules/rxjs/_esm5/index.js`:
- `babel-polyfills` adds `import _Observable from "core-js-pure/features/observable/index.js";` at the top of the file.
- then replaces some (or all) occurences of the term `Observable` with `_Observable`.
- rxjs has a re-export `export { Observable } from './internal/Observable';`. `babel-polyfills` replaces that with `export { _Observable as Observable } from './internal/Observable';`
- bundler rightfully complains that there's no `_Observable` in `./internal/Observable`.

I see two problems here:
1) Why polyfill `Observable` in this case? `Observable` is imported (or re-exported) from somewhere, it's not the global `Observable`.
2) Even if it polyfills, why break the imports?

nicolo-ribaudo:
Fixes #42

### DIFF
--- a/packages/babel-helper-define-polyfill-provider/src/visitors/usage.js
+++ b/packages/babel-helper-define-polyfill-provider/src/visitors/usage.js
@@ -15,6 +15,14 @@ export default (
   return {
     // Symbol(), new Promise
     ReferencedIdentifier(path: NodePath) {
+      if (
+        path.parentPath.isExportSpecifier() &&
+        path.parentPath.parent.source
+      ) {
+        // export { Foo } from "module"
+        return;
+      }
+
       const {
         node: { name },
         scope,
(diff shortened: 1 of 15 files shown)


existentialism (approved):

nicolo-ribaudo:
Actually, this is a bug with `@babel/types`.

nicolo-ribaudo:
Fixed in the main repo: babel/babel#12395

PR #51, opened 2021-01-03
Update corejs3 and its built-in definitions

nicolo-ribaudo:


packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.js
```diff
@@ -185,7 +195,9 @@ export const BuiltIns: ObjectMap<CoreJSPolyfillDescriptor> = {
   compositeKey: define("composite-key", ["esnext.composite-key"]),
   compositeSymbol: define("composite-symbol", ["esnext.composite-symbol"]),
   fetch: define(null, PromiseDependencies),
-  globalThis: define("global-this", ["esnext.global-this"]),
+  globalThis: define("global-this", [
+    { main: "es.global-this", fallbacks: ["esnext.global-this"] },
```
nicolo-ribaudo, line 199: This fallback logic is needed because:
- We can't just use `es.global-this` as the polyfill name to check if it should be included, because it isn't available in older core-js versions
- We can't just use `esnext.global-this`, otherwise we only inject it with `proposals: true` even if not it's stable.

ljharb: Could we just require a new-enough core-js version?

nicolo-ribaudo: Technically yes because this package is still in 0.x so we are free to make breaking changes, but:
- the problem will cam back again in the future, when other proposals are moved to stage 4
- I'd like to keep compatibility with every core-js 3 version, so that we can use these packages in babel/babel#12583

JLHwung: WDYT we add `esnext.some-proposal` as fallback for every `es.some-proposal`? So we don't have to specify main and fallback, instead we can just rename to `es.some-proposal`.

JLHwung (approved):

PR #62, opened 2021-02-22
Use the top-level `targets` option

nicolo-ribaudo:
This feature is for the helper package (not only for corejs3), but I added a test in corejs3 to show the behavior.


packages/babel-plugin-polyfill-corejs3/test/fixtures/top-level-targets/simple/output.mjs
```diff
@@ -0,0 +1,4 @@
+import "core-js/modules/es.promise.finally.js";
+import "core-js/modules/es.promise.js";
+Promise.prototype.finally;
+Array.from;
```
nicolo-ribaudo, line 4: This doesn't polyfill Array.from since it's supported

hzoo (approved):
looks good, other than lint fix

PR #64, opened 2021-02-24
fix: avoid injecting uknown polyfill

clarkdo:
Fix `Internal error in the corejs2 provider: unknown polyfill "web.dom.iterable"`

Related discussion https://github.com/babel/babel/pull/12870#issuecomment-784574841

### DIFF
--- a/packages/babel-plugin-polyfill-corejs2/src/index.js
+++ b/packages/babel-plugin-polyfill-corejs2/src/index.js
@@ -59,7 +59,9 @@ export default defineProvider<Options>(function(
 
   function inject(name: string | string[], utils) {
     if (typeof name === "string") {
-      if (shouldInjectPolyfill(name)) {
+      // Some polyfills aren't always available, for example
+      // web.dom.iterable when targeting node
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
Thanks!

Could you add a test in https://github.com/babel/babel-polyfills/tree/main/packages/babel-plugin-polyfill-corejs2/test/fixtures/usage-global? I think using `Array.from` with `targets: { node: "0.12" }` in the test options should show that the original issue is fixed.

clarkdo:
Test has been added, but the test can be still passing with main branch code, @nicolo-ribaudo did I miss any point ?

nicolo-ribaudo:
I thought it should have failed because  `Array.from` loads `web.dom.iterable` 🤔

clarkdo:
Sorry, it failed indeed, I did't reset to main branch by accident.

image

nicolo-ribaudo (commented):
Thanks!

nicolo-ribaudo (approved):

PR #68, opened 2021-02-26
Remove 1 second delay on process exit

### LINKED ISSUE #63: lodash.debounce for laterLogMissingDependencies delays process exit by 1 second
The `lodash.debounce` for `laterLogMissingDependencies` adds a 1 second delay at process exit while it waits for the `setTimeout` to fire.

https://github.com/babel/babel-polyfills/blob/55f582c8ebc4b8a181a51fecaa92c89158cc7ac9/packages/babel-helper-define-polyfill-provider/src/dependencies.js#L95-L98

This is a really unfortunate side effect of the default `"deferred"` for `missingDependencies.log`. Can we change that default to `"per-file"`?

### Minimal reproducible example

**`babel.config.js`**

```js
module.exports = {
  presets: [["@babel/preset-env", { corejs: 3, useBuiltIns: "entry" }]],
};
```

**`hello.js`**

```js
console.log("Hello, world!");
```

Run

```console
$ yarn add @babel/core @babel/preset-env @babel/register core-js
…
info Direct dependencies
├─ @babel/core@7.13.1
├─ @babel/preset-env@7.13.5
├─ @babel/register@7.13.0
└─ core-js@3.9.0
…
$ export BABEL_DISABLE_CACHE=true  # to reproduce more than once
$ time node -r @babel/register hello.js
Hello, world!

real	0m1.366s
user	0m0.383s
sys	0m0.047s
```

There’s a 1 second delay after printing `Hello, world!` but before exit.

andersk:
Reduce the delay to 100 ms, and don’t delay at all if there’s nothing to log.

Fixes #63.

### DIFF
--- a/packages/babel-helper-define-polyfill-provider/src/node/dependencies.js
+++ b/packages/babel-helper-define-polyfill-provider/src/node/dependencies.js
@@ -95,9 +95,11 @@ let allMissingDeps = new Set();
 const laterLogMissingDependencies = debounce(() => {
   logMissing(allMissingDeps);
   allMissingDeps = new Set();
-}, 1000);
+}, 100);
 
 export function laterLogMissing(missingDeps: Set<string>) {
+  if (missingDeps.size === 0) return;
+
   missingDeps.forEach(name => allMissingDeps.add(name));
   laterLogMissingDependencies();
 }


nicolo-ribaudo (approved):
Thanks!

PR #71, opened 2021-03-02
fix(corejs2): fix polyfills definitions

fangbinwei:
fix definition of `array.every`


nicolo-ribaudo (approved):
Thanks! I wonder how I could make all these errors :sweat_smile:

nicolo-ribaudo:
`babel-plugin-polyfill-corejs2@0.1.10`

PR #74, opened 2021-03-13
Fix typo in documentation (absoluteRuntime -> absoluteImports)

charlessuh:
It's awesome that this works (once I used the correct name.)


nicolo-ribaudo:
Oh thanks 🙃

PR #75, opened 2021-03-13
Provide an ESM build

nicolo-ribaudo:
I'm slowly converting the `babel/babel` repository to ESM, and it's a bit annoying that these packages require the `@babel/core` peerDependency because they fail when testing an ESM build of `@babel/core`.

Providing an ESM build fixes the problem, and it also gives us the possibility of testing real-world ESM Babel plugins.

CI will fail until we release https://github.com/babel/babel/pull/13000.


packages/babel-helper-define-polyfill-provider/package.json
```diff
@@ -19,6 +19,13 @@
     "./lib/node/dependencies.js": "./lib/browser/dependencies.js",
     "./src/node/dependencies.js": "./src/browser/dependencies.js"
   },
+  "exports": [
+    {
+      "import": "./esm/index.mjs",
+      "default": "./lib/index.js"
+    },
+    "./lib/index.js"
+  ],
```
nicolo-ribaudo, line 28: @ljharb Is this correct to ensure compatibility?

Since this package is in 0.x it's ok to disallow imports to other internal files, but I want it to work on every Node.js version since it's what the main Babel packages do.

packages/babel-helper-define-polyfill-provider/src/imports-cache.js
```diff
@@ -1,7 +1,8 @@
 // @flow
 
-import { types as t } from "@babel/core";
 import type { NodePath } from "@babel/traverse";
+import * as babel from "@babel/core";
+const { types: t } = babel.default || babel;
```
nicolo-ribaudo, line 5: This is because Rollup's interop is incompatible with Node.js's behavior (https://github.com/rollup/plugins/issues/635)

test/esm/index.mjs
```diff
@@ -0,0 +1,60 @@
+import assert from "assert/strict";
+import { transformSync } from "@babel/core";
+
+import corejs2 from "babel-plugin-polyfill-corejs2";
+import corejs3 from "babel-plugin-polyfill-corejs3";
+import esShims from "babel-plugin-polyfill-es-shims";
+import regenerator from "babel-plugin-polyfill-regenerator";
+
+assert.equal(typeof corejs2, "function");
```
nicolo-ribaudo, line 9: This would be an object with `{ default: function, __esModule: true }` if we imported the CJS file.

packages/babel-helper-define-polyfill-provider/package.json
ljharb: Yes, exactly! altho i usually do:
```suggestion
  "exports": {
    ".": [
      {
        "import": "./esm/index.mjs",
        "default": "./lib/index.js"
      },
      "./lib/index.js"
    ]
  },
```
I’m not sure if it makes a difference (it shouldn’t, but i haven’t tested the top-level being an array).

however, since you almost certainly want to include “./package.json” here, for tooling, you’d need the dot form anyways, so you could add package.json

nicolo-ribaudo:
I tried using this in an ESM build of the `babel/babel` repository, but it throws because of a Rollup bug: https://github.com/rollup/rollup/issues/3996

Gulpfile.js
```diff
@@ -55,8 +67,57 @@ function build() {
...
+          rollupBabel({
+            envName: "esm",
+            babelrc: false,
+            babelHelpers: "bundled",
+            extends: "./babel.config.js",
+            extensions: [".ts", ".js", ".mjs", ".cjs"],
+          }),
+          rollupNodeResolve({
+            extensions: [".ts", ".js", ".mjs", ".cjs", ".json"],
+            browser: true,
```
JLHwung, line 100: Q: Why do we need to resolve against browser fields?

babel.config.js
```diff
@@ -5,16 +5,7 @@ module.exports = function(api) {
...
-        {
-          loose: true,
-          exclude: ["transform-typeof-symbol"],
-          targets: { node: env === "production" ? "6.9" : "current" },
-        },
-      ],
-      "@babel/preset-flow",
-    ],
+    targets: { node: env === "production" ? "6.9" : "current" },
+    presets: [["@babel/preset-env", { loose: true }], "@babel/preset-flow"],
```
JLHwung, line 9: Not a blocker: let's dogfood the `assumptions`.

JLHwung (approved):

packages/babel-helper-define-polyfill-provider/src/imports-cache.js
nicolo-ribaudo: https://github.com/rollup/plugins/pull/838 will make it possible to avoid doing this, if it will get merged.

Gulpfile.js
nicolo-ribaudo: We now provide two bundles for `@babel/helper-define-polyfill-provider`: one for Node.js (which accesses the filesystem) and one for the browser (which doesn't).

nicolo-ribaudo:
Ohh it's finally green!

PR #78, opened 2021-03-23
[corejs3] Inject `usage-global` imports in the correct order

nicolo-ribaudo:
This PR ensures that we inject dependencies _before_ the polyfill that depends on them. You can use [this 7.12.0 repl](https://babeljs.io/repl/version/7.12.0) to test the old behavior.

Fixes https://github.com/babel/babel-polyfills/issues/77.


zloirock:
See https://github.com/babel/babel/blob/v7.4.0/packages/babel-preset-env/src/polyfills/corejs3/usage-plugin.js#L252

nicolo-ribaudo:
Oh thanks you are right, the best idea is to get the order from `core-js-compat`.

JLHwung (approved):

PR #79, opened 2021-03-23
Fully resolve absolute imports

nicolo-ribaudo:
Node.js doesn't resolve absolute import specifiers, so we have to provide the full path (with the extension, and potentially with the resolved internal file).

Ref: https://github.com/babel/babel/pull/12827


packages/babel-plugin-polyfill-es-shims/test/fixtures/absolute-imports/global/output.mjs
```diff
@@ -1,2 +1,4 @@
-import "<CWD>/array.from/auto";
+import "<CWD>/array.from/auto.js";
+import "<CWD>/math.clz32/auto.js";
 Array.from;
+Math.clz32;
```
nicolo-ribaudo, line 4: `clz32` is different from `.from` because it specifies `package.json#exports`. This test (correctly) shows that it's ignored, because Node.js doesn't consider it when loading absolute paths.

ljharb: Correctly shows that what is ignored? I’m a bit confused here.

ljharb: Also, I’m pretty sure package.json exports aren’t considered by node unless you’re using a bare identifier.

nicolo-ribaudo: The full output is
```js
import "<CWD>/array.from/auto.js";
import "<CWD>/math.clz32/auto.js";
Array.from;
Math.clz32;
```

I added `clz32` to the test to check that it doesn't inject `import "<CWD>/math.clz32/auto"` (because of the `exports` map which includes `"/auto"` and not `"/auto.js"`), since `exports` should be ignored when using absolute paths.

ljharb: ahh ok. why is this preferable over bare identifiers? Absolute paths aren’t portable across machines - doesn’t this increase the likelihood they’ll end up in a published package?

nicolo-ribaudo: > Also, I’m pretty sure package.json exports aren’t considered by node unless you’re using a bare identifier.

Yeah I know, with this test I'm asserting that without "trusting" `require.resolve` (just in case we swap our implementation again in the future, since it's just an internal detail).

nicolo-ribaudo: > ahh ok. why is this preferable over bare identifiers? Absolute paths aren’t portable across machines - doesn’t this increase the likelihood they’ll end up in a published package?

Bare identifiers are preferable, but there are some tools that need absolute paths. I don't recally why, but:
- When you bundle you don't usually care about protability anyway
- We already got different but reports about the missing extension for `@babel/runtime` _when making Babel inject absolute paths_, so it's something that people use.

I don't think people ever used it with `es-shims` imports, but I was fixing it for `regenerator-runtime` and it was easier to fix the general logic than to fix it specifically for a single plugin.

nicolo-ribaudo: Oh for example, next.js uses it (but only for helpers): https://github.com/vercel/next.js/blob/005a8abe39261a8e178ef625ceca9c38ecccaa67/packages/next/build/babel/preset.ts#L187

nicolo-ribaudo: Babel default's is to not use absolute paths. They are enabled by the `absoluteImports`/`absoluteRuntime` options.

ljharb: ahh ok, if someone’s explicitly opted into this then it’s fine ofc

merceyz: > but there are some tools that need absolute paths. I don't recally why

When transpiling stuff in `node_modules` or workspaces while bundling you can't (nor should you) rely on hoisting to give you the correct version so an absolute path is needed.

ljharb: You can indeed rely on that, and using peerDeps, you should.

merceyz: You really can't, If I were to tell next.js to transpile `@material-ui/core` here with the `core-js` version it has access to webpack will throw because it can't find the files in `core-js@2`

```
.
└── node_modules
    ├── next.js
    ├── core-js@3.9
    └── @material-ui/core
        ├── node_modules
        └── core-js@2
```

For your own project sure, you can do it with peer dependencies but that would force all projects using Next.js, `create-react-app`, and `storybook` etc. to explicitly depend on `@babel/runtime`

ljharb: That’s because core-js should be a peer dep of material-ui. I’d file a bug there.

merceyz: `@material-ui/core` doesn't actually depend on `core-js`, I used that name as an example because people usually don't do well with abstracts. If it was a peer dependency then it would be even worse because the root couldn't satisfy both `core-js@3.9` and `core-js@2` at the same time.

ljharb: That's the point - if it peer depends on core-js 2, even implicitly, then it's _incompatible_ with your project and you simply can't use it.

packages/babel-helper-define-polyfill-provider/src/node/dependencies.js
```diff
@@ -18,32 +18,15 @@ export function resolve(
...
-    modulePackage = moduleName;
-    moduleNestedPath = "";
-  } else {
-    modulePackage = moduleName.slice(0, slash);
-    moduleNestedPath = moduleName.slice(slash);
-  }
-
   try {
-    let pkg;
     if (nativeRequireResolve) {
```
JLHwung, line 22: Not related to this PR: we need to update `nativeRequireResolve` to use cheap semver so versions like `8.10` does not fall to `requireResolve`.

JLHwung (approved):

PR #85, opened 2021-04-16
Fix shim/polyfill terminology

ljharb:
This follows the terms used in the [es-shims API](https://github.com/es-shims/es-shim-api#api-contract).

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,9 @@ For example, if you want to inject imports to [`es-shims`](https://github.com/es
 <td>
 
 ```js
- 
- 
- 
+
+
+
 Promise.allSettled([
   p1,
   p2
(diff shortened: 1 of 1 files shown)


zloirock:
I don't like es-shims terminology. Please, don't use the term "shim" here.

ljharb:
Can you elaborate on why not? The term "shim" is pretty universally used in the ecosystem, predating even the es5-shim.

zloirock:
For a long time for almost all "polyfill" at first means self-installed polyfill with feature detection, "shim" is not a word usually used for that. I don't think that it's good to concentrate attention on it. Now it's a word that could confuse more than "polyfill".

nicolo-ribaudo:
Can we use the terms "self-installing polyfill" and "pure polyfill", specifying each time where it's needed?

ljharb:
Sure, I’m fine with that. Will update shortly.

nicolo-ribaudo (approved):
Thanks!

PR #93, opened 2021-07-07
[corejs3] Fix instance.at proposals imports

essentia0:
Hey there !

PR code is pretty much explanatory, but it fixes the missing imports from the [relative indexing methods proposal](https://github.com/tc39/proposal-relative-indexing-method/) aka "String.prototype.at", "Array.prototype.at", etc.

Currently only `esnext.string.at` is imported, which causes problems when using the array (or typed-array) version.

I focused on adding tests for `usage-global`, because of the `include` [issue](https://github.com/babel/babel/issues/8932) which complicates a workaround. Let me know if you need additional tests.


zloirock (approved):

nicolo-ribaudo (commented):
Thanks!

PR #94, opened 2021-07-19
[corejs3,es-shims] Add array "find from last" methods

nicolo-ribaudo:


JLHwung (approved):
We should update https://babeljs.io/docs/en/babel-preset-env#shippedproposals, too.

ljharb (approved):

PR #98, opened 2021-08-24
Use `es.*` for stable esnext features

### LINKED ISSUE #97: (corejs3) `"".replaceAll` not being polyfilled in Node v14 LTS
Please see https://github.com/zloirock/core-js/issues/978 for details on the problem I am seeing (I think there is value in seeing the response given by the maintainer, so that is why I haven't copied the issue contents over here). As mentioned in that issue, please let me know if you cannot reproduce this yourself, and I'll try to provide a minimal case. Let me know if you have any other questions!

JLHwung:
Fixes #97 

Added a `builtin-definition-checker` so we don't forget to update it anymore.


ljharb:
Could we add a checker for es-shims also?

JLHwung:
@ljharb Sure, will do it in a separate PR.

zloirock (commented):
IIRC logic of falling back `es.` -> `esnext.` had some problems, could you add tests for that?

packages/babel-plugin-polyfill-corejs3/test/fixtures/usage-pure/built-in-instance-methods/output.js
```diff
@@ -128,7 +132,7 @@ _reduceRightInstanceProperty(object);
 
 _repeatInstanceProperty(object);
 
-object.replaceAll;
```
JLHwung, line 131: Since the `replaceAll` is stage 4, we should polyfill it if users are using an older version of corejs which provides `esnext.string.replace-all`.

nicolo-ribaudo (approved):

PR #103, opened 2021-09-10
[es-shims] Add support for the `String.prototype` Annex B methods

nicolo-ribaudo:


packages/babel-plugin-polyfill-es-shims/data/polyfills.json
```diff
@@ -824,5 +823,174 @@
...
+    "firefox": "17",
+    "safari": "6",
+    "node": "0.10",
+    "android": "4",
+    "ios": "7",
+    "phantom": "2",
+    "samsung": "1",
+    "electron": "0.20"
+  },
+  "String.prototype.sup": {
```
ljharb, line 983: I believe all of these work in ie 10 and 11 (but haven’t checked lately)

ljharb (approved):

packages/babel-plugin-polyfill-es-shims/data/polyfills.json
nicolo-ribaudo: I don't have access to IE to test it, but this data is generated from `compat-table`.

ljharb: Oops, I’ll update the compat table then, please hold :-)

ljharb: The data in the compat table under "name: 'String.prototype HTML methods'" includes `ie10: true,` for one test, and `ie11: false,` for another two which means all IE versions require the polyfill. What am i missing?

nicolo-ribaudo: It might be a bug in the script which extracts the data, let me check :thinking:

nicolo-ribaudo: Ahhh, this file contains _supported_ versions. IE is not marked as supported, so it always include the polyfill.

nicolo-ribaudo: I added a test

ljharb: ahhh ok thanks, that wasn't clear to me

PR #116, opened 2022-03-01
docs: remove leftover references to duplicate targets

### LINKED ISSUE #18: Mention the rfc to move targets top level in the readme?
Using this makes it clear we don't want to have different sets of target options https://github.com/babel/rfcs/pull/2

```
{
  "presets": [
    ["@babel/preset-env", {
      "targets": ">1%, not ie 11"
    }]
  ],
"plugins": [
    [
      "babel-plugin-polyfill-corejs3",
      {
        "method": "usage-global",
        "targets": {
          "edge": 16
        }
      }
    ]
  ]
}
```

agilgur5:
## Summary

Follow-up to 1091fa4e27a597148403545d21e776dadeb5c518

## Details

- 1091fa4e27a597148403545d21e776dadeb5c518 switched the examples to use
  top-level targets, but this note at the top was leftover

- in the same commit, the comments within JSON were removed, so the
  syntax highlighting can be set back to regular JSON instead of
  JSON5 or JSONC
  - Notably, GitHub was not correctly rendering anything other than JSON

- also fix syntax highlighting within an HTML table by using
  HTML `<code>` blocks instead of Markdown backticks
  
### Screenshot

Here's what the improperly highlighted sections looked like on GitHub (both the JSON5 and the backticks):
Screen Shot 2022-03-01 at 6 44 10 PM

Can see the before/after below:
[Old Markdown Preview](https://github.com/babel/babel-polyfills/blob/1091fa4e27a597148403545d21e776dadeb5c518/docs/migration.md#es-shims) vs. [New Markdown Preview](https://github.com/babel/babel-polyfills/blob/1ec29ee32752e9b382235d979d1d71fd3b0bbbb9/docs/migration.md#es-shims)

## References

Closes #18  
It's been open for a while and this removes the last reference to the top-level targets RFC, so it is now a quite outdated issue

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -2,8 +2,6 @@
 
 All the existig capabilities of `@babel/preset-env` and `@babel/plugin-transform-runtime` related to `core-js` are supported by the polyfill providers implemented in this repository.
 
-> You'll notice that you currently have to duplicate your `targets` (unles you are using a `.browserslistrc` file). We are working on deduplicating the option: [`babel/rfcs#2`](https://github.com/babel/rfcs/pull/2).
-
 ### `core-js@3`
 
 <!-- prettier-ignore-start -->
@@ -117,7 +115,7 @@ All the existig capabilities of `@babel/preset-env` and `@babel/plugin-transform
 </td>
 <td>
 
-```json5
+```json
 {
   "presets": [
     ["@babel/preset-env", {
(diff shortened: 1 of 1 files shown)


agilgur5:
@nicolo-ribaudo could you take a look at this PR too? this one's been waiting around a bit longer

nicolo-ribaudo (approved):
I forgot about this PR, thanks!

PR #119, opened 2022-03-19
feat: throw error when targets are used with regenerator

### LINKED ISSUE #36: polyfill-regenerator doesn't support targets
## Issue

Hey there, so I've been trying to add this to [TSDX](https://github.com/formium/tsdx) in https://github.com/formium/tsdx/pull/795 but ran into an issue: https://github.com/formium/tsdx/pull/795#issuecomment-694697396. Long story short, `polyfill-regenerator` doesn't seem to support `targets`.

If I'm understanding correctly, [this line](https://github.com/babel/babel-polyfills/blob/3bb29d0b7c18e1baf9227481c03bff20a5854860/packages/babel-plugin-polyfill-regenerator/src/index.js#L9) needs to specify `targets` per [the `polyfill-provider` docs](
https://github.com/babel/babel-polyfills/blob/main/docs/polyfill-provider.md#providerpolyfills-string---name-string-support-).

Would be happy to write a PR for this and will look into it more tomorrow. I'm not sure if I should just manually input [the data from MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*#Browser_compatibility) or if I should use `babel-compat-data` [for `transform-regenerator`](https://github.com/babel/babel/blob/44f8287d7ac3c523a9425a7bb4beb4882090ae20/packages/babel-compat-data/data/plugins.json#L372)? Any guidance would be appreciated

## Context

Per above, was looking to get this to work with [TSDX](https://github.com/formium/tsdx), which would add quite a lot of indirect library users of this.

Concretely, I was adding this to replace the now defunct/unmaintained [`babel-plugin-async-to-promises`](https://github.com/rpetrich/babel-plugin-transform-async-to-promises) (that has some [correctness](https://github.com/formium/tsdx/labels/topic%2Fasync-to-promises) [bugs](https://github.com/rpetrich/babel-plugin-transform-async-to-promises/issues)) without the downsides of an impure `preset-env` `useBuiltIns` or all-or-nothing `transform-runtime`.

## P.S.

Also just wanted to say really happy to see this set of libraries existing to provide a solution for gaps between `preset-env` `useBuiltIns` and `transform-runtime`, solves a long existing problem!

agilgur5:
## Summary

adds an error when `polyfill-regenerator` is supplied with `targets`, as those are unsupported for this polyfill per #36

## Details

- the `polyfill-regenerator` plugin actually does not support targets and
  does not use them
  - it instead detects whether the `regeneratorRuntime` variable has
    been injected by `transform-regenerator`, which already supports
    targets
  - so targets are not needed twice in that sense

- instead of silently ignoring the targets option, error out when it is
  given as it is an unsupported option
  - this should hopefully make it easier for developers to realize
    what's going on when debugging issues with this (as I did)

- add a test for this new error as well as a basic "harness" based on
  [the one in `define-polyfill-provider`](https://github.com/babel/babel-polyfills/blob/dfc74d8a78cb3f655dacf8aceea13a2a6024dee3/packages/babel-helper-define-polyfill-provider/test/options.js)
  - remove `targets` from any fixtures that had them for this polyfill
    so that this test passes
    - and because they didn't do anything previously anyway
 
 ## Review Notes

- I had to create a basic test "structure"/"harness" to check for this error, so let me know if that needs stylistic/structural modification. I tried copying `define-polyfill-provider`'s error testing with a similar [`options.js` test file](https://github.com/babel/babel-polyfills/blob/dfc74d8a78cb3f655dacf8aceea13a2a6024dee3/packages/babel-helper-define-polyfill-provider/test/options.js)
- Wording maybe isn't the best, so let me know if you'd like any changes to the error message
 
## References

Fixes #36


packages/babel-plugin-polyfill-regenerator/test/options.js
```diff
@@ -0,0 +1,20 @@
...
+describe("targets", () => {
+  it("is not supported", () => {
+    expect(() =>
+      transform("code", {
+        method: "usage-pure",
+        targets: { chrome: 50 },
+      }),
+    ).toThrow(/targets/);
+  });
+});
```
nicolo-ribaudo, line 20: Can you also add a test that makes sure that top-level targets are supported? i.e.
```json5
{
  targets: { chrome: 50 },
  plugins: [polyfillRefenerator, { method: "usage-pure" }]
}
```
should not throw (because the top-level targets may be useful to other plugins).

agilgur5: Good idea to double-check that!

agilgur5: Hmm... that was a good catch, it does seem to (incorrectly) throw when top-level targets are used.

I'm actually not sure myself how to differentiate between top-level and plugin option within the arguments here...

Is there a way to differentiate simply that you know of? Or would that require some non-trivial refactoring to even implement?
I have a feeling it might be the latter 😅

agilgur5: Added the failing tests here in a new commit for reference

nicolo-ribaudo:
@agilgur5 I was thinking that we could just completely remove the `targets` options from these plugins: users can specify it as a top-level option.

We would still need a way to pass it as an internal API to the `corejs2` and `corejs3` plugins (with an "hidden" option similar to `"#__secret_key__@babel/runtime__compatibility"`) for compatibility with `@babel/preset-env` until Babel 8, but there is no other need to support the `targets` plugin option now that we have top-level targets.

nicolo-ribaudo:
As a shorter term solution (that one is a bigger breakig change), we can modify the check like this:
```js
export default defineProvider(({ debug, targets, babel }, options) => {
  if (!shallowEqual(targets, babel.targets())) {
    throw new Error(
      "This plugin does not use the targets option. Only preset-env's targets" +
        " or top-level targets need to be configured for this plugin to work." +
        " See https://github.com/babel/babel-polyfills/issues/36 for more" +
        " details",
    );
  }
```
where `shallowEqual` checks if they have the same properties with the same values.

This doesn't catch the case where you explicitly specify the same top-level targets and plugin targets, but it catches the "common error" which is to specify only plugin-level targets (thus not using top-level targets).

packages/babel-plugin-polyfill-regenerator/test/options.js
agilgur5: Added the fix as described [below](https://github.com/babel/babel-polyfills/pull/119#issuecomment-1076893006)

packages/babel-plugin-polyfill-regenerator/src/index.ts
```diff
@@ -39,3 +48,7 @@ export default defineProvider<Options>(({ debug }, options) => {
 
 const isRegenerator = meta =>
   meta.kind === "global" && meta.name === "regeneratorRuntime";
+
+function shallowEqual(obj1: any, obj2: any) {
+  return JSON.stringify(obj1) === JSON.stringify(obj2);
```
agilgur5, line 53: This is obviously a pretty naive implementation of shallow equality, but it has no deps and is _very_ short. Let me know if that's okay or if you'd prefer something less naive

agilgur5 (commented):
Rebased & updated with latest changes and added the shallow equality fix for top-level targets. Tests now pass!

packages/babel-plugin-polyfill-regenerator/src/index.ts
nicolo-ribaudo: I don't like it, but it's ok in this case since it's comparing very small serializable objects.

packages/babel-plugin-polyfill-regenerator/src/index.ts
```diff
@@ -8,7 +8,16 @@ type Options = {
...
 };
 
-export default defineProvider<Options>(({ debug }, options) => {
+export default defineProvider<Options>(({ debug, targets, babel }, options) => {
+  if (!shallowEqual(targets, babel.targets())) {
+    throw new Error(
+      "This plugin does not use the targets option. Only preset-env's targets" +
+        " or top-level targets need to be configured for this plugin to work." +
+        " See https://github.com/babel/babel-polyfills/issues/36 for more" +
+        " details",
```
nicolo-ribaudo, line 17:
```suggestion
        " details.",
```

nicolo-ribaudo (approved):

nicolo-ribaudo:
Thanks!

packages/babel-plugin-polyfill-regenerator/src/index.ts
agilgur5: Yep that was kind of my thought process too 👍

PR #120, opened 2022-03-26
deps: upgrade to Prettier v2

agilgur5:
## Summary

upgrade Prettier to v2 to support newer TS features once the repo is migrated to TS

- first step in TS migration and currently required for `flowts`

## Details 

- to support newer TS features once the repo is migrated to TS
  - namely `import type`

- rename [deprecated `jsxBracketSameLine` config to new `bracketSameLine`](https://prettier.io/blog/2021/09/09/2.4.0.html)

- some formatting was updated as a result:
  - `function(` -> `function (`
  - some random spacing things:
    - helper's `index.js` call to `instantiateProvider` was reformatted
    - corejs3's `built-in-definition.js` `StaticProperties` had a
      large reformatting

## Review Notes

Can add a [`.git-blame-ignore-revs`](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) similar to [Babel core](https://github.com/babel/babel/blob/e703e9f838459068bdc4e6b4ef119649862bbbba/.git-blame-ignore-revs) for this change in another PR once this is merged (if wanted, of course)

Next steps for migration are already almost complete, see PR #121


nicolo-ribaudo (approved):
Thanks!

I pushed a commit to minimize the diff; it's now small enough that we don't need `.git-blame-ignore-revs`

agilgur5:
Ah great reduction!

PR #121, opened 2022-03-26
refactor: flow -> ts

agilgur5:
## Summary

Converts JS/Flow files to TS with [`flowts`](https://github.com/zxbodya/flowts) and _many_ manual interventions.
Adds support for TS in linting, formatting, building, and CI.

## TODO

- [x] Merge #120 first
- [x] Fix remaining type errors
- [x] Remove `$FlowIgnore` or replace with `@ts-expect-error` where still necessary
- [x] Remove Flow entirely~?~
  - ~I think this should be fine but not 100% sure that nothing uses it anymore~ -- could not find other references to Flow (e.g. with `// @flow` comments), so removed it
  - [x] Remove from `package.json`, `yarn.lock`, `babel.config.json`, `.flowconfig`, `ci.yml`, `package.json:scripts`, `CONTRIBUTING.md`
- [x] ~Convert tests to TS?~ No, see below
  - `flowts` seemed to leave all of these alone. Fixtures should remain as JS, but test files themselves could be converted.  Babel core seemed to still have tests in JS as well unless I missed something
    - Yep, there is [a perf rationale](https://github.com/babel/babel-polyfills/pull/121#issuecomment-1086963899) for Babel tests to be in JS 
- [x] ~add `.git-blame-ignore-revs` on top of the `flowts` commit?~ No, this [won't work anyway](https://github.com/babel/babel-polyfills/pull/121#issuecomment-1086951241)
  - namely for the [`types.js` big change](https://github.com/babel/babel-polyfills/pull/121#discussion_r835837502) due to `,` -> `;` change

## Details

1. add `tsconfig.json` and `typescript` dep

1. support linting and formatting TS files
    - add `@typescript-eslint` for linting with type information
    - add TS extensions to ESLint config and Prettier config
    - fix `no-unused-vars` with [custom rule](https://github.com/babel/babel-polyfills/pull/121#discussion_r835836191) similar to [Babel core](https://github.com/babel/babel/blob/e703e9f838459068bdc4e6b4ef119649862bbbba/.eslintrc.cjs#L41)
  
1. support building TS files
    - add `@babel/preset-typescript` to parse and build TS files similar to [Babel core](https://github.com/babel/babel/blob/e703e9f838459068bdc4e6b4ef119649862bbbba/babel.config.js#L164)
    - modify `Gulpfile` + Rollup config to handle TS similar to [Babel core](https://github.com/babel/babel/blob/e703e9f838459068bdc4e6b4ef119649862bbbba/Gulpfile.mjs#L370)
    - modify `shipped-proposals` script to output as TS
 
1. auto convert Flow -> TS via flowts
    - some manual changes/adjustments:
      - manually reversed some incorrect formatting changes (lines with `// eslint-ignore` or `// prettier-ignore`)
      - manually reversed various unnecessary spacing / new line changes (as [requested by Nicolò](https://github.com/babel/babel-polyfills/pull/121#discussion_r835882722))
      - manually [renamed](https://github.com/babel/babel-polyfills/pull/121#discussion_r835836627) some auto-named vars
  
    - major changes seem to be:
      - removing `// @flow` at the top of files
      - `*` -> `any`
      - `Object` -> `any`
      - `?type` -> `type | undefined | null`
      - using semicolons instead of commas for object property types
   
    - 453 type errors post-conversion
 
1. manually remediate several type errors post-conversion
    - went from 453 errors at the beginning to 40 errors!
    
    - some missing types:
      - missing [generics](https://github.com/babel/babel-polyfills/pull/121#discussion_r835837180) in a few places, particularly with Sets (e.g. `Set<string>`)
      - missing [optional](https://github.com/babel/babel-polyfills/pull/121#discussion_r835836801) parameters in a few places (e.g. `subfolder?`)
      - missing [default values](https://github.com/babel/babel-polyfills/pull/121#discussion_r835851945) in some initializers (e.g. `{ useBabelRuntime = "" }`)
    
    - some fixes:
      - add [`| typeof presetEnvSilentDebugHeader`](https://github.com/babel/babel-polyfills/pull/121#discussion_r836070601) to fix debug checks
      - add [`"object" in meta` check](https://github.com/babel/babel-polyfills/pull/121#discussion_r835883699) to `usageGlobal` funcs to narrow type before assuming `meta.object` exists
      - one NodePath type import was incorrect / [different from the rest](https://github.com/babel/babel-polyfills/pull/121#discussion_r835836457)
      - one `hasOwnProperty` was called as a global func instead of on `Object` or `Object.prototype`

    - remove `babel.default || babel` imports
        - tsconfig `esModuleInterop` should handle this
    
1. remove or replace `$FlowIgnore` with `@ts-expect-error`
    - in most places, there was no type error, so just removed the `// $FlowIgnore` comment
    - in 2 places, there was a type error, so replace with a `// @ts-expect-error` comment instead
      - this could be removed with some type narrowing checks instead, but this is a more direct conversion (and requested in review comments [below](https://github.com/babel/babel-polyfills/pull/121#discussion_r835883791))

    - 38 type errors remaining

1.  refine non-trivial types to remediate remaining type errors
    
    - add `@types/babel__traverse` as a devDep for more accurate typings
      - this was [already a transitive dep](https://github.com/babel/babel-polyfills/pull/121#discussion_r836078911), so adding it just updated it to a newer (and more accurate) version
    
    - some complicated generics refined:
      - missing `Options` in regenerator's `defineProvider` call
      - many missing generics for various `NodePath` args
    
    - had to switch up some [`isXXX()` conditionals](https://github.com/babel/babel-polyfills/pull/121#pullrequestreview-923386800) and surroundings to narrow the right type
      - unfortunately the type narrowing doesn't work on every property or to the entire depth of the tree 😕 
    
    - add a `@ts-expect-error` where proper type resolution was not possible with the current Babel types
    - use casts where TS was not able to narrow the type all the way

1. support TS type-checking in CI
    - add `yarn tscheck` command (same naming as [Babel core](https://github.com/babel/babel/blob/e703e9f838459068bdc4e6b4ef119649862bbbba/Makefile#L79))
    - modify `yarn validate` to also run `yarn tscheck`
    - modify `ci.yml` GitHub workflow to also run `yarn tscheck`
    - add `yarn tscheck` to `CONTRIBUTING.md`

1. remove all Flow deps, scripts, config, CI, etc 
    - remove Flow dep and `.flowconfig`
    - remove Flow scripts from `package.json` and `ci.yml`
    - remove Babel config and Babel plugin
    - ESLint plugin is required by `eslint-config-babel`, so cannot be removed
  
## Review Notes

- Can see each individual commit for more details and a more detailed commit description
- I found https://github.com/babel/babel/pull/12317 -- let me know if you'd like me to split more of the "prep" parts as separate PRs (all the `env` commits). I wasn't sure since you had said "single PR" in https://github.com/babel/babel-polyfills/pull/118#issuecomment-1075588382
- Still some TODOs left as per above
  
## References

- See thread at https://github.com/babel/babel-polyfills/pull/118#pullrequestreview-915099428 that kicked off the conversion. It references multiple things from Babel core's Flow -> TS migration:
  - https://github.com/babel/babel/pull/11578
  - https://github.com/babel/babel/pull/12317
  - ["Flow -> TS" label](https://github.com/babel/babel/labels/Flow%20-%3E%20TS)
- #120 and #122 (ish) were split off from this


packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
```diff
@@ -1,8 +1,5 @@
-// @flow
-
 import type { NodePath } from "@babel/traverse";
-import * as babel from "@babel/core";
-const { types: t } = babel.default || babel;
+import { types as t } from "@babel/core";
```
agilgur5, line 2: so the previous version was erroring in TS.

per my commit message, `esModuleInterop` in `tsconfig.json` should normally handle this automatically... but we're building with `@babel/preset-typescript`, not TS, which, if I recall correctly, does not use `tsconfig.json`, so I'm not sure if this holds entirely

packages/babel-plugin-polyfill-es-shims/src/index.ts
```diff
@@ -30,8 +28,8 @@ export default defineProvider<{||}>(function({
   });
 
   function createDescIterator(
-    cb: (Descriptor, Utils, Object) => void,
-    instance?: (MetaDescriptor, *, Utils, Object) => void,
+    cb: (c: Descriptor, b: Utils, a: any) => void,
+    instance?: (d: MetaDescriptor, c: any, b: Utils, a: any) => void,
```
agilgur5, line 32: this is a `flowts` thing -- let me know if there's any specific var names you'd prefer instead

packages/babel-plugin-polyfill-es-shims/src/mappings.ts
```diff
@@ -162,7 +158,7 @@ for (const name of [
   });
 }
 
-function createDescriptor(name, version, pkg = name.toLowerCase(), subfolder) {
+function createDescriptor(name, version, pkg = name.toLowerCase(), subfolder?) {
```
agilgur5, line 161: manually had to make the last param optional in a few funcs below because it's not always passed in its usage. not sure why Flow hadn't errored on these before?

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "lib": [
+      "esnext"
+    ],
+    "noEmit": true,
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "isolatedModules": true,
+    "skipLibCheck": true,
+    "resolveJsonModule": true,
+    "paths": {
```
agilgur5, line 15: fixed the duplicated `compilerOptions` in https://github.com/babel/babel-polyfills/pull/118#issuecomment-1073346266

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "include": ["./packages/*/src/**/*.ts"],
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "esnext",
+    "lib": [
+      "esnext"
+    ],
+    "noEmit": true,
```
agilgur5, line 9: would you like to add declarations and declaration maps later? or do you not want to publish types with these packages for now?

I noticed [Babel core had both](https://github.com/babel/babel/blob/e703e9f838459068bdc4e6b4ef119649862bbbba/tsconfig.base.json#L9), and I generally recommend using both. Not sure if your intent with `noEmit` was just to simplify things for now or something else

packages/babel-plugin-polyfill-corejs2/src/built-in-definitions.ts
```diff
@@ -1,36 +1,37 @@
...
 const pureAndGlobal = (
   pure: string,
   global: string[],
-  minRuntimeVersion: ?string = null,
+  minRuntimeVersion: string | undefined | null = null,
 ) => define<CoreJS2Meta>(global[0], pure, global, { minRuntimeVersion });
 
-const globalOnly = (global: string[]) => define(global[0], null, global);
+const globalOnly = (global: string[]) =>
+  define<CoreJS2Meta>(global[0], null, global);
```
agilgur5, line 34: missing generic here caused a TS error. guess it doesn't error in Flow for some reason?

packages/babel-helper-define-polyfill-provider/src/index.ts
```diff
@@ -41,13 +39,16 @@ function resolveOptions<Options>(
...
-  providerOptions: ProviderOptions<Options>,
-  absoluteImports: string | boolean,
+  method: MethodString;
+  methodName: "usageGlobal" | "entryGlobal" | "usagePure";
+  targets: Targets;
+  debug: boolean;
+  shouldInjectPolyfill:
+    | ((name: string, shouldInject: boolean) => boolean)
+    | undefined
+    | null;
```
agilgur5, line 49: there are some [functional differences between](https://github.com/niieani/typescript-vs-flowtype#maybe--nullable-type) Flow's Maybe type and TS's optional type, so, per my commit message, you'll see similar changes in several places

packages/babel-helper-define-polyfill-provider/src/types.ts
```diff
@@ -0,0 +1,120 @@
+import type { NodePath } from "@babel/traverse";
+import { types as t } from "@babel/core";
+import createMetaResolver from "./meta-resolver";
+
+type ObjectMap<T> = {
+  [k: string]: T;
```
agilgur5, line 6: per my commit message, this file got counted as a deletion + addition instead of a rename bc TS uses semicolons at the end of property types, while Flow uses commas. as this file is entirely types, that meant most of the file got reformatted

packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.ts
```diff
@@ -20,7 +20,7 @@ const define = (
   pure,
   global,
   name = global[0],
-  exclude,
+  exclude?,
```
agilgur5, line 23: similar to the other file, this threw a TS error as it wasn't optional before, but is not always used

packages/babel-helper-define-polyfill-provider/src/types.ts
```diff
@@ -0,0 +1,120 @@
...
+  absoluteImports?: boolean | string;
+};
+
+export type PolyfillProvider<Opts = {}> = (
+  api: ProviderApi,
+  options: ProviderOptions<Opts>,
+  dirname: string,
+) => ProviderResult;
+
+export type PolyfillProviderInternal<Opts> = PolyfillProvider<Opts>;
```
agilgur5, line 47: this was previously `export type PolyfillProviderInternal = PolyfillProvider<Opts>;` which gave a bunch of TS errors but somehow was valid Flow? afaict, should be a generic on both sides

agilgur5 (commented):
left some in-line comments for clarity/reference as well as some questions

packages/babel-plugin-polyfill-corejs2/src/index.ts
```diff
@@ -11,27 +9,25 @@ import addPlatformSpecificPolyfills from "./add-platform-specific-polyfills";
...
+    runtimeVersion: string;
+    ext: string;
+  };
+};
 
-export default defineProvider<Options>(function(
+export default defineProvider<Options>(function (
   api,
   {
     [presetEnvCompat]: { entryInjectRegenerator } = {},
```
agilgur5, line 33: this line and several similar lines give the type error: `error TS2525: Initializer provides no value for this binding element and the binding element has no default value`

adding a default value to each key fixes this, but I'm not sure if that's entirely the proper solution

packages/babel-plugin-polyfill-corejs3/src/index.ts
```diff
@@ -291,7 +288,7 @@ export default defineProvider<Options>(function(
           resolved.desc,
           resolved.name,
           utils,
-          // $FlowIgnore
+          // @ts-ignore
           meta.object,
```
agilgur5, line 292: this statement and the similar one below it could be refactored with some type narrowing to remove the `// @ts-ignore`, e.g.:

```suggestion
          "object" in meta ? meta.object : undefined,
```

agilgur5 (commented):
Investigated some of the type errors some more!

Left a few in-line comments on the remaining type errors.

Other than those, the rest are of the form, e.g. `error TS2339: Property 'value' does not exist on type 'Node'. Property 'value' does not exist on type 'AnyTypeAnnotation'.` -- basically the type has not been narrowed enough.
- This could be solved with some narrowing checks, typecasts, `// @ts-ignore`, etc. Or may be due to incomplete type info somewhere?

agilgur5:
I'm also not entirely sure if the snapshot changes are correct or not... I actually hadn't seen them locally until recently.

Investigated that as I thought they might have came from my manual type remediations, but actually they seem to begin with the `flowts` commit 🤔 
I probably missed them initially because you need to run `yarn build` in this repo before running `yarn test`, otherwise you end up testing stale code, which is probably what I had done locally while developing.

package.json
```diff
@@ -9,25 +9,29 @@
...
     "test:esm": "node test/esm/index.mjs",
-    "validate": "yarn lint && yarn flow && yarn test",
+    "tscheck": "yarn tsc",
+    "validate": "yarn lint && yarn flow && yarn tscheck && yarn test",
     "watch": "gulp watch"
   },
   "devDependencies": {
     "@babel/cli": "^7.13.10",
-    "@babel/core": "^7.13.10",
+    "@babel/core": "^7.17.8",
```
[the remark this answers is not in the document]
nicolo-ribaudo: I updated all of them in https://github.com/babel/babel-polyfills/pull/122 :+1:

.eslintrc.json
```diff
@@ -9,7 +9,16 @@
   "env": { "node": true },
   "overrides": [
     {
-      "files": ["packages/*/src/**/*.js"],
+      "files": ["**/*.ts"],
+      "parser": "@typescript-eslint/parser",
+      "plugins": ["@typescript-eslint"],
+      "rules": {
+        "no-unused-vars": "off",
+        "@typescript-eslint/no-unused-vars": "error"
```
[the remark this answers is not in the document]
nicolo-ribaudo: This is enough! We'll add the other overrides only when we are affected by the relative invalid errors.

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
nicolo-ribaudo: The problem is that we are compiling this file both to CJS and to ESM: when it runs directly in Node.js, the import is only accessible as `babel.default.types`. The best thing is probably to do something like this to make every tool happy:
```js
const { types: t } = (babel.default || babel) as typeof import("@babel/core");
```
and stick a `@ts-expect-error` in front of it if TS complains.

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
```diff
@@ -1,8 +1,5 @@
-// @flow
-
 import type { NodePath } from "@babel/traverse";
```
nicolo-ribaudo, line 1: We might have to add `@types/babel__traverse` to this package's `devDependencies`

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
```diff
@@ -48,7 +45,10 @@ export default class ImportsCache {
       source: t.StringLiteral,
       // eslint-disable-next-line no-undef
       name: t.Identifier,
-    ) => { node: t.Node, name: string },
+    ) => {
+      node: t.Node;
+      name: string;
+    },
```
nicolo-ribaudo, line 51:
```suggestion
    ) => { node: t.Node; name: string },
```

I know that this is an unnecessary nitpick and that it's done by flowts, but unnecessary newlines in small objects annoy me :sweat_smile:

packages/babel-helper-define-polyfill-provider/src/index.ts
```diff
@@ -41,13 +39,16 @@ function resolveOptions<Options>(
...
-  methodName: "usageGlobal" | "entryGlobal" | "usagePure",
-  targets: Targets,
-  debug: boolean,
-  shouldInjectPolyfill: ?(name: string, shouldInject: boolean) => boolean,
-  providerOptions: ProviderOptions<Options>,
-  absoluteImports: string | boolean,
+  method: MethodString;
+  methodName: "usageGlobal" | "entryGlobal" | "usagePure";
+  targets: Targets;
+  debug: boolean;
```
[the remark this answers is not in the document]
nicolo-ribaudo: We can mark that magic string as const:
```ts
export const presetEnvSilentDebugHeader =
  "#__secret_key__@babel/preset-env__don't_log_debug_header_and_resolved_targets" as const;
```
and then annotate this as
```ts
debug: boolean | typeof presetEnvSilentDebugHeader;
```

packages/babel-helper-define-polyfill-provider/src/index.ts
nicolo-ribaudo: While `?` means "or null or undefined", I believe in all the occurrences in this repo `| undefined` should be enough (unless TS complains obviously)

packages/babel-helper-define-polyfill-provider/src/types.ts
```diff
@@ -0,0 +1,120 @@
...
+export type ProviderOptions<Opts = {}> = {
+  include?: Pattern[];
+  exclude?: Pattern[];
+} & Opts;
+
+export type PluginOptions = {
+  method: MethodString;
+  targets?: {
+    browsers: string | string[];
+    [target: string]: string | number;
```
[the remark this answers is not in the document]
nicolo-ribaudo: This is correct, because you can have things like `node: 16` or `node: "16.7"`.

packages/babel-helper-define-polyfill-provider/src/utils.ts
```diff
@@ -1,18 +1,15 @@
-// @flow
-
-import * as babel from "@babel/core";
-const { types: t, template } = babel.default || babel;
-import type NodePath from "@babel/traverse";
+import { types as t, template } from "@babel/core";
+import type { NodePath } from "@babel/traverse";
```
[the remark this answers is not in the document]
nicolo-ribaudo: We might have to add `@types/babel__traverse` to this package's `devDependencies`, becuase the packages published from the babel/babel repo don't include yet type definitions (except for `@babel/types` and `@babel/parser`). We'll start publishing them in Babel 8.

packages/babel-plugin-polyfill-corejs2/src/index.ts
nicolo-ribaudo: This error is actually correct, the default value should have been
```js
[presetEnvCompat]: { entryInjectRegenerator } = { entryInjectRegenerator: false },
```

packages/babel-plugin-polyfill-corejs2/src/index.ts
```diff
@@ -116,6 +112,7 @@ export default defineProvider<Options>(function(
 
       if (
         resolved.kind !== "global" &&
+        "object" in meta &&
         meta.object &&
```
nicolo-ribaudo, line 116: This is something from TS I hate :sob: I guess I'll have to accept it.

packages/babel-plugin-polyfill-corejs3/src/index.ts
nicolo-ribaudo: It's ok with the suppression comment (I actually prefer it). However, it's better to use `@ts-expect-error` so that if one day the TS compiler becomes smart enough that it's not needed we get "notified" that the comment is not needed anymore.

packages/babel-plugin-polyfill-es-shims/src/index.ts
nicolo-ribaudo:
```suggestion
    cb: (desc: Descriptor, utils: Utils, path: NodePath) => void,
    instance?: (meta: MetaDescriptor, resolved: any, utils: Utils, path: NodePath) => void,
```

tsconfig.json
nicolo-ribaudo: I prefer not to publish the type definitions yet; we'll do it when releasing Babel 8 for the main packages.

nicolo-ribaudo:
The failing tests are because of the Babel update; just rebase on `main` and the failures should go away.

packages/babel-plugin-polyfill-corejs2/src/index.ts
agilgur5: It makes sense at least! 
JS is just a wonky language where you can do `meta.object` and if it doesn't exist it's just `undefined`. Would throw a `KeyError` in Python.

So I like to think of it as TS "fixing" some JS if that reframing helps with your acceptance 😅

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
agilgur5: I think it's done by the Prettier portion of `flowts`, but I'll try changing it to this and see if Prettier complains

packages/babel-helper-define-polyfill-provider/src/types.ts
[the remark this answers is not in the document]
agilgur5: I meant that `browsers` and `target` seem to be incompatible with each other (removing one or the other resolves the type error).

The type error (can see in CI logs) is basically complaining that `string[]` is not an `index` type

packages/babel-plugin-polyfill-corejs3/src/index.ts
agilgur5: Ah I forgot about that feature!
Though I'm not sure if TS would ever try changing that since it's a safety thing that JS doesn't have

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
agilgur5: Or not, guess it was a `recast` thing or something? guess Prettier has no opinion there, so I was able to revert it without issues. Same for the newlines added `ObjectMap` in `corejs3/src/built-in-definitions.ts`

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
agilgur5: I see now that the unit tests pass that `test:esm` fails due to this. 

When compiling with Babel, is there no equivalent of TS's `esModuleInterop` (and the whole `__esModule` property etc) that would make it compatible with both without such a hack?

test/esm/package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "name": "@babel-internal/test-esm",
   "devDependencies": {
-    "@babel/core": "^7.13.10",
+    "@babel/core": "^7.17.8",
```
agilgur5, line 5: I guess this somehow didn't get updated in #122 ? I believe there's a similar line in the `yarn.lock` from this

packages/babel-helper-define-polyfill-provider/src/index.ts
agilgur5: Gotcha. That might take me some time to change all the occurrences to TS's optional type `?` and remove the `| undefined | null`

packages/babel-helper-define-polyfill-provider/src/utils.ts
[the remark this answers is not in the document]
agilgur5: So when I did a "Go to Definition" in VS Code for `NodePath`, it did take me to `@types/babel__traverse` in `node_modules`. So I believe it may already be there due to a transitive dependency.

I did try `yarn add -dev @types/babel__traverse` and that only fixed 3 type errors 😕

agilgur5 (commented):
> The failing tests are because of the Babel update

Ah right, I updated that later while developing and not during the initial conversions, so that explains it! Fixed now after rebase.

Fixed several comments and type errors, but unfortunately still 38 type errors left (and +10 if `createDescIterator` changes are made), almost entirely stemming from `NodePath` errors.
Per in-line comment, adding `@types/babel__traverse` only fixed 3 type errors 😕

nicolo-ribaudo (commented):
A few pointers to fix some of the TS errors:

- `packages/babel-helper-define-polyfill-provider/src/imports-cache.ts`
  - The various `programPath` parameters should be `NodePath<t.Program>`. Probably also a few weakmaps need to be updated
  - The code in `_injectImport` is horribly written, probably we should split the usage of the `lastImport` variable in two variables (`lastImport` and `newNodes`)
- `packages/babel-helper-define-polyfill-provider/src/utils.ts`
  - The `path` param of `resolveKey` should be a `NodePath<t.Expression>`
  - Using `t.isXXX(node)` instead of `path.isXXX()` in `resolveKey` will probably help with type refinements
  - You might need a few `as NodePath<t.Expression>` after some `.get("...")` calls
  - `getImportSource`'s parameter should be a `NodePath<t.ImportDeclaration>`
  - I don't think `._blockHoist` is in the type definitions; just put a `@ts-expect-error` in front of it
- `packages/babel-helper-define-polyfill-provider/src/visitors/usage.ts`
  - All the visitor functions need a better `NodePath` annotation, for example `MemberExpression(path: NodePath)` should be `MemberExpression(path: NodePath<t.MemberExpression>)`. The node for `ReferencedIdentifier` is just `t.Identifier`

packages/babel-plugin-polyfill-corejs2/src/index.ts
```diff
@@ -11,31 +9,33 @@ import addPlatformSpecificPolyfills from "./add-platform-specific-polyfills";
...
 
 export default defineProvider<Options>(function (
   api,
   {
-    [presetEnvCompat]: { entryInjectRegenerator } = {},
-    [runtimeCompat]: { useBabelRuntime, runtimeVersion, ext = ".js" } = {},
+    [presetEnvCompat]: { entryInjectRegenerator = false } = {},
+    [runtimeCompat]: {
+      useBabelRuntime = "",
+      runtimeVersion = "",
```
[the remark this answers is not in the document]
nicolo-ribaudo: Those defaults are ok. However, if the object is provided it should already have all the properties and not rely on a default fallback. We should define it like this:

```js
[presetEnvCompat]: { entryInjectRegenerator } = { entryInjectRegenerator: false }
```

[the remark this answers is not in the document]
agilgur5: Is there a functional difference between the two? not sure if I'm missing something or if that's just stylistic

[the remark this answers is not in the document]
nicolo-ribaudo: There is a difference when you pass an empty object:
```js
function a({ x } = { x: 2 }) { return x }
function b({ x = 2 } = {}) { return x }

a({}); // undefined
b({}); // 2
```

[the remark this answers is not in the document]
nicolo-ribaudo: In this specific case, I'm asking for the first behavior because you should either pass a complete object with all the necessary properties, or you don't pass the object. You cannot pass a "partial object", where some properties are specified and others have the default value.

agilgur5 (commented):
~15ish type errors left 😅  

Getting there. Think I might be able to get another half of them, but it's getting trickier with each refinement.

The tips helped a lot! Also the new type info made it easier for me to understand the AST parsing going on as well 🙂

packages/babel-plugin-polyfill-corejs2/src/index.ts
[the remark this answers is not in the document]
agilgur5: Ah I see, thanks for the clarification! The destructuring on destructuring syntax got a little confusing for me.

I'll convert it to this format shortly. Should I convert the `ext = ".js"` default as well? As that one was already there

packages/babel-plugin-polyfill-es-shims/src/index.ts
agilgur5: I renamed everything but left out the `NodePath` refinement for now

packages/babel-helper-define-polyfill-provider/src/utils.ts
```diff
@@ -98,19 +100,21 @@ export function getRequireSource({ node }: NodePath) {
     t.isCallExpression(expression) &&
     t.isIdentifier(expression.callee) &&
     expression.callee.name === "require" &&
-    expression.arguments.length === 1 &&
-    t.isStringLiteral(expression.arguments[0]);
-  if (isRequire) return expression.arguments[0].value;
+    expression.arguments.length === 1;
+  if (isRequire && t.isStringLiteral(expression.arguments[0])) {
+    return expression.arguments[0].value;
+  }
```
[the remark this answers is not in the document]
agilgur5: rewrote it a bit for readability

packages/babel-plugin-polyfill-corejs2/src/index.ts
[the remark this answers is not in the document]
agilgur5: Converted to this format, but left `ext = ".js"` as that's what it was before. Let me know if `ext` should be changed as well

packages/babel-helper-define-polyfill-provider/src/types.ts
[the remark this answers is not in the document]
agilgur5: For this to type-check, it would have to be:

```suggestion
    browsers: string | string[];
    [target: string]: string | string[] | number;
```

Since TS requires the index signature [to be a union](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures) of all the existing types

That actually changes the code, so I'm not really sure what to change this to as I'm not sure what was meant to be expressed

packages/babel-plugin-polyfill-corejs3/src/index.ts
```diff
@@ -214,21 +212,17 @@ export default defineProvider<Options>(function (
...
         if (!path.isMemberExpression()) return;
         if (!path.isReferenced()) return;
 
-        isCall = path.parentPath.isCallExpression({ callee: path.node });
-
         if (meta.key === "Symbol.iterator") {
           if (!shouldInjectPolyfill("es.symbol.iterator")) return;
 
-          if (isCall) {
+          if (t.isCallExpression(path.parent, { callee: path.node })) {
```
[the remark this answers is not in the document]
agilgur5: rewrote this a bit to use the original call (`path.parentPath.isCallExpression`), but same issue exists there

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
```diff
@@ -76,12 +82,12 @@ export default class ImportsCache {
...
-      lastImport = lastImport.insertAfter(node);
+      newNodes = lastImport.insertAfter(node);
     } else {
-      lastImport = programPath.unshiftContainer("body", node);
+      newNodes = programPath.unshiftContainer("body", node);
     }
-    lastImport = lastImport[lastImport.length - 1];
-    this._lastImports.set(programPath, lastImport);
+    const newNode = newNodes[newNodes.length - 1];
+    this._lastImports.set(programPath, newNode);
```
agilgur5, line 90: modified this function [as requested](https://github.com/babel/babel-polyfills/pull/121#pullrequestreview-923386800), but this last line still has a type error (`Node` vs. `Program`)

packages/babel-helper-define-polyfill-provider/src/utils.ts
```diff
@@ -87,30 +91,33 @@ export function resolveSource(obj: NodePath) {
   return { id: null, placement: null };
 }
 
-export function getImportSource({ node }: NodePath) {
+export function getImportSource({ node }: NodePath<t.ImportDeclaration>) {
   if (node.specifiers.length === 0) return node.source.value;
 }
 
-export function getRequireSource({ node }: NodePath) {
+export function getRequireSource({ node }: NodePath<t.Statement>) {
```
agilgur5, line 98: not sure if this generic is correct; this was like the least common denominator

packages/babel-helper-define-polyfill-provider/src/types.ts
```diff
@@ -0,0 +1,108 @@
...
+  include?: Pattern[];
+  exclude?: Pattern[];
+} & Opts;
+
+export type PluginOptions = {
+  method: MethodString;
+  targets?: {
+    browsers: string | string[];
+    [target: string]: string | string[] | number;
+  };
```
agilgur5, line 28: modified this [as mentioned](https://github.com/babel/babel-polyfills/pull/121#discussion_r838015118), but per that discussion, not sure if this is the intended type

agilgur5 (commented):
Ok, got to 9 type errors now (+7 if `NodePath` is added to `createDescIterator` in `es-shims`).
I've also added more types in various places where I inferred what it should be based on the new/existing patterns of typings (went from like ~200 type errors if you turn on `strict` to ~130 type errors)

Don't think I can get any farther without more "hints" as I don't actually know what the code _should_ be.

Did some trial-and-error here and there that you might have seen in the intermediary pushes (e.g. adding too many generics?). The types in `imports-cache` were particularly confusing and onerous to wrap my brain around, but got through all but one of them

packages/babel-helper-define-polyfill-provider/src/types.ts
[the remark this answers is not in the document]
agilgur5: See follow-up discussion in https://github.com/babel/babel-polyfills/pull/121#discussion_r838033963

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
nicolo-ribaudo: Actually, the error is correct and we should change the `_lastImport` type to `WeakMap<NodePath<t.Program>, NodePath<t.Node>>` (technically it's `WeakMap<NodePath<t.Program>, NodePath<t.ImportDeclaration | t.ImportStatement>>`, but it would require a bunch of type casts).

packages/babel-helper-define-polyfill-provider/src/utils.ts
```diff
@@ -34,12 +31,16 @@ function resolveId(path) {
...
-  const { name } = node;
+export function resolveKey(
+  path: NodePath<t.Expression>,
+  computed: boolean = false,
+) {
+  const { scope } = path;
+  if (path.isStringLiteral()) return path.node.value;
   const isIdentifier = path.isIdentifier();
-  if (isIdentifier && !(computed || parent.computed)) return name;
+  if (isIdentifier && !(computed || path.parent.computed)) {
```
nicolo-ribaudo, line 41: We can type cast `path.parent` to a `t.MemberExpression`.

packages/babel-helper-define-polyfill-provider/src/visitors/usage.ts
```diff
@@ -24,7 +23,7 @@ export default (
       callProvider({ kind: "global", name }, path);
     },
 
-    MemberExpression(path: NodePath) {
+    MemberExpression(path: NodePath<t.MemberExpression>) {
       const key = resolveKey(path.get("property"), path.node.computed);
       if (!key || key === "prototype") return;
```
nicolo-ribaudo, line 29: (Line 31-32 below)

the type error is correct, we should only check the binding if `object.isIdentifier()`.

packages/babel-helper-define-polyfill-provider/src/utils.ts
```diff
@@ -34,12 +31,16 @@ function resolveId(path) {
   }
 }
 
-export function resolveKey(path: NodePath, computed: boolean = false) {
-  const { node, parent, scope } = path;
-  if (path.isStringLiteral()) return node.value;
-  const { name } = node;
+export function resolveKey(
+  path: NodePath<t.Expression>,
```
nicolo-ribaudo, line 35:
```suggestion
  path: NodePath<t.Expression | t.PrivateName>,
```
(the type errors about this are correct)

packages/babel-plugin-polyfill-corejs2/src/index.ts
```diff
@@ -154,7 +157,7 @@ export default defineProvider<Options>(function (
           meta.key === "Symbol.iterator" &&
```
nicolo-ribaudo, line 157: (Line 141 above)

 We can cast `path.node` to `BinaryExpression`, thanks to the check at line 133. Also in the corejs3 plugin.

nicolo-ribaudo (commented):
Some more hint I could find looking at the ci TS errors.

In the next few days I will clone this PR locally to debug the remaining failures!

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
agilgur5: `ImportStatement` didn't seem to be an available type. I tried `Import` and `ImportSpecifier` though, and they only errored on this same line, i.e. didn't introduce more errors as far as I could tell.

Changed it to `t.Node` for now though

packages/babel-helper-define-polyfill-provider/src/visitors/usage.ts
agilgur5: added another conditional check for this

packages/babel-plugin-polyfill-corejs2/src/index.ts
agilgur5: Ah, I figured one of the conditionals would _implicitly_ narrow it, but did not know how to interpret that myself. Both have been cast now

packages/babel-helper-define-polyfill-provider/src/utils.ts
agilgur5: this one narrowed `path` correctly, but `path.parent` was still just a `Node`. Cast it to a `t.MemberExpression` now

packages/babel-plugin-polyfill-es-shims/src/index.ts
```diff
@@ -30,10 +29,15 @@ export default defineProvider<{||}>(function ({
...
 
   function createDescIterator(
-    cb: (Descriptor, Utils, Object) => void,
-    instance?: (MetaDescriptor, *, Utils, Object) => void,
+    cb: (desc: Descriptor, utils: Utils, path: any) => void,
+    instance?: (
+      meta: MetaDescriptor,
+      resolved: any,
+      utils: Utils,
+      path: any,
```
agilgur5, line 37: Per [prev discussion](https://github.com/babel/babel-polyfills/pull/121#discussion_r837907615), I left this as `any` for now as changing it to just `NodePath` (with no generic or further specification) adds ~7 type errors

packages/babel-plugin-polyfill-corejs3/src/index.ts
```diff
@@ -54,7 +52,7 @@ export default defineProvider<Options>(function (
     version = 3,
     proposals,
     shippedProposals,
-    [runtimeCompat]: { useBabelRuntime, ext = ".js" } = {},
+    [runtimeCompat]: { useBabelRuntime, ext = ".js" } = { useBabelRuntime: "" },
```
agilgur5, line 55: Per [prev discussion](https://github.com/babel/babel-polyfills/pull/121#discussion_r837995855), left `ext = ".js"` in in this file and the other as it was already there. Let me know if you'd like me to change this

packages/babel-plugin-polyfill-corejs3/src/index.ts
```diff
@@ -300,12 +294,12 @@ export default defineProvider<Options>(function (
...
+          // @ts-expect-error
           meta.object,
         );
         if (!id) return;
 
-        if (isCall) {
+        if (path.parentPath.isCallExpression({ callee: path.node })) {
           callMethod(path, id);
         } else {
           path.replaceWith(t.callExpression(id, [path.node.object]));
```
agilgur5, line 305: This is the _very last_ type error currently -- I assume it's because TS isn't able to properly narrow on `else`s or something

agilgur5 (commented):
We now have just 1 more type error remaining!

@nicolo-ribaudo could you go through the leftover unresolved discussions and either comment or react or resolve them? If I received a thumbs-up on some of my comments, I resolved them myself, but some got no reaction so I wasn't sure if you had seen them or not.
I made a list below to make it easier for you to check things off.

### Leftover Unresolved Discussions

- These were just informational comments, I think they can be resolved unless you have thoughts otherwise:
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r835836801
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r835837180
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r835837502
    - Per my PR description, not sure if a `.git-blame-ignore-revs` might be worthwhile here?
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r836051399
    - Not sure if there's anything to do here

- Some annoying changes that I couldn't workaround but are equivalent expressions:
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r837995556
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r838017716
  
- Some types I wasn't 100% sure of:
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r836822349
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r838031151
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r838033963

- [x] [CJS/ESM compat issues](https://github.com/babel/babel-polyfills/pull/121#discussion_r840877843) that still aren't quite functioning

- The 3 in-line comments below
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r840848747 
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r840851685
  - [x] https://github.com/babel/babel-polyfills/pull/121#discussion_r840852267

### Remaining TODOs for me

- [ ] Convert tests to TS? Per my PR description, `flowts` left them all alone, but I think they might in JS in Babel core too
- [ ] Split things into different PRs? (per PR description)
- [x] [Flow's Maybe type refinement](https://github.com/babel/babel-polyfills/pull/121#discussion_r836077257)

packages/babel-plugin-polyfill-es-shims/src/index.ts
agilgur5: See [follow-up discussion](https://github.com/babel/babel-polyfills/pull/121#discussion_r840848747)

packages/babel-plugin-polyfill-corejs2/src/index.ts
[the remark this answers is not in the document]
agilgur5: See [follow-up discussion](https://github.com/babel/babel-polyfills/pull/121#discussion_r840851685)

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
agilgur5: This would require a `// @ts-expect-error` everywhere as it does still error due to "Property 'default' does not exist on type... "

The cast does fix `t`'s type so that it doesn't interpret as `any` due to that error.

nicolo-ribaudo:
Oh thanks for that comment with all the things I didn't reply to :sweat_smile: 

> `.git-blame-ignore-revs`

I don't think this will work anyway, because the file has also been renamed and git thinks that it's a new file, so git doesn't now how to track the original contents.

(I'm still going through the other checkboxes)

packages/babel-helper-define-polyfill-provider/src/utils.ts
nicolo-ribaudo: Yup, that's correct. It could even be `Node` since it's refined on the line after, but `Statement` is more precise.

packages/babel-helper-define-polyfill-provider/src/types.ts
nicolo-ribaudo: TS complains but then it seems to correctly pick up the types:

TS playground screenshot

I'll just put a `@ts-expect-error` to suppress it.

packages/babel-plugin-polyfill-corejs3/src/index.ts
nicolo-ribaudo: This is all legacy code that hopefully will be removed at some point in the future; let's leave this as-is :grimacing:

nicolo-ribaudo:
> Convert tests to TS? Per my PR description, flowts left them all alone, but I think they might in JS in Babel core too

I'd prefer to leave them as JS, so that in the future we can run them without on-the-fly transpilation and use github.com/nicolo-ribaudo/jest-light-runner (which I just extracted from the Babel repo) because it has better support for ESM compared to Jest.

> Split things into different PRs? (per PR description)

Nah, this is still small enough that it's not needed.

---

This _should be ready_ now, if I didn't miss something feel free to mark it as "ready" so that I can then merge it :slightly_smiling_face:

packages/babel-helper-define-polyfill-provider/src/types.ts
agilgur5: Weird 👀  👀 , I'm surprised that works. I wonder if the unsound-ness might affect some other type resolution 🤔

packages/babel-helper-define-polyfill-provider/src/normalize-options.ts
```diff
@@ -5,7 +5,7 @@ import type {
   MissingDependenciesOption,
 } from "./types";
 
-function patternToRegExp(pattern: Pattern): RegExp | undefined | null {
+function patternToRegExp(pattern: Pattern): RegExp | null {
```
agilgur5, line 8: Thanks for handling the Flow Maybe type refinement!

Just wanted to confirm for this line and a few others -- was removing `undefined` intentional as opposed to removing `null`? Just thought I'd double check in code review since you had said [the reverse](https://github.com/babel/babel-polyfills/pull/121#discussion_r835882979) before

scripts/babel-plugin-compat-core-import.js
```diff
@@ -0,0 +1,31 @@
+// Convert
```
agilgur5, line 1: I thought this was just some Node script initially, then I realized it was a tiny Babel plugin and I was like 😮 😮 😮 

Guess I should expect no less of one of Babel's maintainers to just be able to write a Babel plugin so quickly on the fly like that 😅 

Great fix!

agilgur5 (commented):
Just one comment, otherwise I think it's good to go!

Thanks for sticking through this process with me @nicolo-ribaudo ! This was actually my first time touching any actual _parser_ code in any Babel plugin (as opposed to everything else). Hopefully this experience makes it easier for me to write parser plugins myself in the future 🙂 

Always a good experience interacting with you Nicolò ❤️

packages/babel-helper-define-polyfill-provider/src/index.ts
agilgur5: Completed by Nicolò in https://github.com/babel/babel-polyfills/commit/80605cc0084c4adfd6b50c2d626ce2bd169a4236 . See follow-up in https://github.com/babel/babel-polyfills/pull/121#discussion_r845379333

packages/babel-helper-define-polyfill-provider/src/imports-cache.ts
agilgur5: Nicolò wrote up a tiny Babel plugin to workaround this in https://github.com/babel/babel-polyfills/pull/121/commits/68f27965a2d2d3779bda0f8d1e33b68346c291b4

agilgur5:
> Oh thanks for that comment with all the things I didn't reply to 😅

Glad to hear that the list I made was useful! 🙂 

> I don't think this will work anyway, because the file has also been renamed and git thinks that it's a new file, so git doesn't now how to track the original contents.

Ah, right I forgot it got counted an addition/deletion so it wouldn't be tracked as a same file change as such 😕  Might be caught by a different diff algorithm but don't think GitHub supports those yet. Good catch!

> I'd prefer to leave them as JS, so that in the future we can run them without on-the-fly transpilation and use github.com/nicolo-ribaudo/jest-light-runner (which I just extracted from the Babel repo) because it has better support for ESM compared to Jest.

Thanks for the context! I figured there must be a reason for that, [perf differential](https://github.com/babel/babel/pull/13966#pullrequestreview-819765720) makes sense. 
That being said, I've sometimes found having types in my tests to be useful for finding some errors, and [type tests](https://github.com/jaredpalmer/tsdx/issues/781) are a thing too. But probably more useful for Babel 8+

packages/babel-helper-define-polyfill-provider/src/normalize-options.ts
nicolo-ribaudo: Yeah, it looks like in many places the code explicitly passes `null` to mean "no value", rather than just not passing/returning anything.

scripts/babel-plugin-compat-core-import.js
nicolo-ribaudo: Sometimes you don't have to look to far to find the right tool for the job!

agilgur5: > Sometimes you don't have to look to far to find the right tool for the job!

Hahaha very true!

> I use ASTExplorer to write these tiny plugins

Ooooh that makes a lot of sense. I've seen AST Explorer before, but in context while writing a plugin that is _significantly_ more useful 👀 👀 👀 
Most plugins I'd been thinking of writing would be tiny, for instance, like [literate programming](https://github.com/babel/babel/issues/2592) support (though this would require a [custom parser](https://babeljs.io/docs/en/babel-parser#will-the-babel-parser-support-a-plugin-system) now that I remember where I left off) or [`tsconfig` paths](https://github.com/jaredpalmer/tsdx/issues/91#issuecomment-679541162) support, so that's a really useful tool! Thanks for the example use-case!

PR #126, opened 2022-04-10
[es-shims] Add change-array-by-copy proposal methods

nicolo-ribaudo:
I also added `findLast`/`findLastIndex` to the readme, because they have been supported since #94.

I didn't add the group-by proposal yet, to wait for the new name.

### DIFF
--- a/scripts/build-es-shims-data/index.js
+++ b/scripts/build-es-shims-data/index.js
@@ -5,6 +5,19 @@ const path = require("path");
 const { generateData, environments, writeFile } = require("./utils-build-data");
 
 const newData = generateData(environments, require(`./es-shims-features`));
+
+// These features are missing from compat-table. Remove this from this list once they are added.
+const missing = [
+  "Array.prototype.toReversed",
+  "Array.prototype.toSorted",
+  "Array.prototype.toSpliced",
+  "Array.prototype.with",
+];
+for (const name of missing) {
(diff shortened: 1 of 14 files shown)


scripts/build-es-shims-data/index.js
```diff
@@ -5,6 +5,19 @@ const path = require("path");
...
 
 const newData = generateData(environments, require(`./es-shims-features`));
+
+// These features are missing from compat-table. Remove this from this list once they are added.
+const missing = [
+  "Array.prototype.toReversed",
+  "Array.prototype.toSorted",
+  "Array.prototype.toSpliced",
+  "Array.prototype.with",
+];
```
nicolo-ribaudo, line 15: `compat-table` does not have entries for this proposal yet.

ljharb: That’s because they aren’t shipped anywhere, so everything needs them. Perhaps it should just assume that when data is missing?

ljharb (approved):

JLHwung (approved):

scripts/build-es-shims-data/index.js
nicolo-ribaudo: Yes, the currently have `{}` as "support data" so they are always included. I added them to this list because the current build process warns when using a feature not present in compat-table (so that we don't forget about wiring them).

PR #129, opened 2022-05-04
[es-shims] Update compat data and add `parseInt`

nicolo-ribaudo:
@ljharb I didn't add the new `.splice`, `.push` and `.concat` polyfills because I couldn't find entries for those methods in compat-table (the compat data in this repository are automatically generated from compat-table, so that we don't have to maintain yet another source).


ljharb (approved):
There’s more cases for parseint; whitespace isn’t handled properly around the Mongolian vowel separator in a bunch of node versions. I’ll try to update the compat table for all these things.

JLHwung:
> whitespace isn’t handled properly around the Mongolian vowel separator

This is somehow expected, the general category of U+180E MVS was changed from Zs to Cf in 2012 for Unicode 4.0.
Later it was changed from Cf to Zs for Unicode 6.3. Therefore certain Node versions aligning to Unicode 4.0 - 6.2 would recognize U+180E as a whitespace.

ljharb:
@JLHwung what i would expect is that in any engine where `/\s/.test('\u180e')` returns `true`, then `parseInt` would ignore that character as leading whitespace (and if `false`, not) but there are a number of engines where this was not the case.

PR #131, opened 2022-06-02
[es-shims] Add Date and Array#unshift polyfills

nicolo-ribaudo:
cc @ljharb 

For most of the Date polyfills I couldn't find any entry in compat-table.


ljharb:
I'll work on adding some Date tests to compat-table, but in one way or another, i believe every browser needs *something* from the package :-)

PR #138, opened 2022-07-31
Update `core-js`

zloirock:


zloirock:
@nicolo-ribaudo

JLHwung (dismissed):

packages/babel-plugin-polyfill-corejs3/src/shipped-proposals.ts
```diff
@@ -1,8 +1,10 @@
 // This file is automatically generated by scripts/build-corejs3-shipped-proposals.mjs
 
 export default new Set<string>([
-  "esnext.array.find-last",
-  "esnext.array.find-last-index",
-  "esnext.typed-array.find-last",
-  "esnext.typed-array.find-last-index",
+  "esnext.array.to-reversed",
```
nicolo-ribaudo, line 4: Is it already available in some browsers by default?

zloirock: It's generated automatically, so yes it's available in some *environments* by default. It's not a browser, it's `bun`.

JLHwung: But `preset-env`'s shippedProposals does not include `bun`. Until `preset-env` recognizes `bun` as shippedProposals, can we exclude `bun` here?

https://github.com/babel/babel-polyfills/blob/ce30946033aaed0ae1daf09a75e76acf4d24b8cc/scripts/build-corejs3-shipped-proposals.mjs#L21

zloirock: I think that it should be solved on the `preset-env` side, https://github.com/babel/babel/issues/14831.

zloirock: So, this proposal is supported in Safari 16 that's released today. If you wanna change something in the `shippedProposals` logic - feel free to do it out of the scope of this PR - I think that now it's incorrect anyway.

nicolo-ribaudo:
Merged in https://github.com/babel/babel-polyfills/commit/1f2008f0111af5ad5b1ffc0cf7dbcc5ceba9ad18

zloirock:
@nicolo-ribaudo why I wrote that the way of considering something as "shipped" is incorrect? It's not even the definition of "shipped".

Safari 16 implements the "change array by copy" proposal - all actual methods. However, recently was changed the order of errors in `%TypedArray%.prototype.with`. Because of this minor issue, `core-js-compat` shows that this method is not supported in Safari 16. With the logic of "shipped proposals" list generation, to this will be added all methods from this proposal with the exception of `esnext.typed-array.with`.

It should be added too.

nicolo-ribaudo:
@zloirock I'm happy to delegate the decision of what counts as shipped to you: do you know where we can get the "is this supported enough, even if with bugs, to consider it as shipped" data?

zloirock:
@zloirock at this moment we could only manually support this list.

nicolo-ribaudo:
Note that apparently I forgot to run `yarn build` when merging this PR; the result is now this: https://github.com/babel/babel-polyfills/blob/main/packages/babel-plugin-polyfill-corejs3/src/shipped-proposals.ts

zloirock:
Yep, since it's fixed in Safari TP / Bun.

PR #140, opened 2022-08-19
fix: avoid transpile super call

sanyuan0704:
Avoid super call transformation which is unespected:
```ts
class List {
  map() {
    return;
  }
}

class TreeNodeList extends List {
  constructor() {
   // In current version
   // Compiled to `_mapInstanceProperty(super)`
    super.map;
   // Compiled to _mapInstanceProperty(_context = super).call(_context);
    super.map();
  }
}
```
Instead, babel polyfill plugin should skip the case.

### DIFF
--- a/packages/babel-plugin-polyfill-corejs3/src/index.ts
+++ b/packages/babel-plugin-polyfill-corejs3/src/index.ts
@@ -219,6 +219,9 @@ export default defineProvider<Options>(function (
         if (!path.isMemberExpression()) return;
         if (!path.isReferenced()) return;
         if (path.parentPath.isUpdateExpression()) return;
+        if (path.isMemberExpression() && t.isSuper(path.node.object)) {
+          return;
+        }
 
         if (meta.key === "Symbol.iterator") {
           if (!shouldInjectPolyfill("es.symbol.iterator")) return;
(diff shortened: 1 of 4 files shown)


zloirock (changes requested):
Instead of `List` here can be, for example, `Iterator` or `AsyncIterator` and `.map` should be polyfilled for them.

sanyuan0704:
> Instead of `List` here can be, for example, `Iterator` or `AsyncIterator` and `.map` should be polyfilled for them.

Okay, i just give a example.In fact, the super call transformation is unspectedly.

zloirock:
@nicolo-ribaudo I don't think that here is possible to handle getters properly without rewriting all `core-js` entries for instance methods. It's one of many limitations of the pure version.

I can add `receiver` argument to those entries, but it will be available only from a new minor version and overcomplicate the logic - it's better to think about for a new major version.

I don't know what's better - proper getters of properly polyfilled parent methods.

packages/babel-plugin-polyfill-corejs3/src/index.ts
```diff
@@ -219,6 +219,9 @@ export default defineProvider<Options>(function (
         if (!path.isMemberExpression()) return;
         if (!path.isReferenced()) return;
         if (path.parentPath.isUpdateExpression()) return;
+        if (path.isMemberExpression() && t.isSuper(path.node.object)) {
```
JLHwung, line 222:
```suggestion
        if (t.isSuper(path.node.object)) {
```
`path` must be a member expression at this point.

JLHwung (approved):
I think it is good to merge it as-is. Before this PR, we are generating invalid code. After this PR the usage is not polyfilled, not ideal but at least the output work on modern environments. We can figure out how to handle `super` later.

liuxingbaoyu (approved):

PR #143, opened 2022-09-12
Improve absolute paths on windows.

liuxingbaoyu:
Used to fix babel8 test exceptions on windows.


nicolo-ribaudo:
Thank you! CI failure is fixed on main.

Do you want to also open a PR adding windows to CI?

liuxingbaoyu:
There seems to be some issues on Windows, I'll try that.

PR #154, opened 2023-02-02
fix: Some dependencies were incorrectly excluded

liuxingbaoyu:
I'm not sure this is a good approach, since it assumes that `array` and `string` do not depend on each other. (but this shouldn't break anything new, since it's more permissive than the original filtering)
And `"iterator", "async-iterator", "dom-collections"` should be safe, because we can only detect `iterator`.

Fixes: https://github.com/babel/babel/issues/15392


nicolo-ribaudo (approved):
This seems to be ok :+1:

PR #156, opened 2023-02-16
Update `core-js`

zloirock:


packages/babel-plugin-polyfill-corejs3/src/shipped-proposals.ts
```diff
@@ -4,13 +4,6 @@ export default new Set<string>([
   "esnext.array.from-async",
   "esnext.array.group",
   "esnext.array.group-to-map",
-  "esnext.array.to-reversed",
-  "esnext.array.to-sorted",
-  "esnext.array.to-spliced",
-  "esnext.array.with",
   "esnext.string.is-well-formed",
   "esnext.string.to-well-formed",
```
zloirock, line 8: Note that it can be removed by automatic build because of a bug in WebKit - but it's still a shipped proposal.

JLHwung (approved):

packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.ts
```diff
@@ -682,6 +704,7 @@ export const InstanceProperties = {
   codePoints: define("instance/code-points", ["esnext.string.code-points"]),
   concat: define("instance/concat", ["es.array.concat"], undefined, ["String"]),
   copyWithin: define("instance/copy-within", ["es.array.copy-within"]),
+  demethodize: define("instance/demethodize", ["esnext.function.demethodize"]),
```
nicolo-ribaudo, line 707: I have a question. This proposal has never been presented to TC39, what's the core-js policy regarding when to implement a proposal? If a TC39 delegate creates a repository for it?

zloirock: It was presented in the scope of another proposal, https://github.com/js-choi/proposal-function-demethodize/blob/main/HISTORY.md - it was decided to extract it.

zloirock: > If a TC39 delegate creates a repository for it?

https://github.com/js-choi/proposal-function-demethodize

ljharb: Be forewarned that this name isn’t likely to last.

zloirock: Agree. But in any case, it's better than outdated.

PR #158, opened 2023-02-26
Update `core-js`

zloirock:


packages/babel-plugin-polyfill-corejs3/src/shipped-proposals.ts
```diff
@@ -4,6 +4,12 @@ export default new Set<string>([
...
   "esnext.array.group",
   "esnext.array.group-to-map",
+  "esnext.set.intersection.v2",
+  "esnext.set.is-disjoint-from.v2",
+  "esnext.set.is-subset-of.v2",
+  "esnext.set.is-superset-of.v2",
+  "esnext.set.symmetric-difference.v2",
+  "esnext.set.union.v2",
   "esnext.string.is-well-formed",
   "esnext.string.to-well-formed",
```
zloirock, line 14: Note that now it's fixed in WebKit, so the build script does not break this list.

JLHwung (approved):

PR #160, opened 2023-03-30
Implement new transform-runtime protocol for runtime package name

nicolo-ribaudo:
The goal of this new API is that any polyfill provider can specify their own polyfilled version of `@babel/runtime` (such as `@babel/runtime-corejs3`), instead of hard-coding `@babel/runtime`, `@babel/runtime-corejs2` and `@babel/runtime-corejs3` in `@babel/plugin-transform-runtime`.

This is done by returning a `runtimeName: string` property in the polyfill provider result object, which will be picked up by `@babel/plugin-transform-runtime` when injecting helpers.

See https://github.com/babel/babel/pull/15531 for the other part of this change. This PR needs to land first.

### DIFF
--- a/packages/babel-helper-define-polyfill-provider/src/index.ts
+++ b/packages/babel-helper-define-polyfill-provider/src/index.ts
@@ -321,11 +321,29 @@ export default function definePolyfillProvider<Options>(
       console.log(`\nUsing polyfills with \`${method}\` method:`);
     }
 
+    const { runtimeName } = provider;
+
     return {
       name: "inject-polyfills",
       visitor,
 
-      pre() {
+      pre(file) {
+        if (runtimeName) {
+          if (
+            file.get("runtimeHelpersModuleName") &&
+            file.get("runtimeHelpersModuleName") !== runtimeName
+          ) {
+            console.warn(
+              `Two different polyfill providers are trying to define two` +
+                ` conflicting @babel/runtime alternatives:` +
+                ` ${file.get("runtimeHelpersModuleName")} and ${runtimeName}.` +
+                ` The second one will be ignored.`,
+            );
+          } else {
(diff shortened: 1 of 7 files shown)


packages/babel-plugin-polyfill-corejs2/src/index.ts
```diff
@@ -22,7 +24,7 @@ type Options = {
     entryInjectRegenerator: boolean;
   };
   "#__secret_key__@babel/runtime__compatibility": void | {
-    useBabelRuntime: string;
+    useBabelRuntime: boolean;
```
JLHwung, line 27: Is this a breaking change? If so this PR should be landed in a new major.

nicolo-ribaudo: Yes, but these packages are still at 0.x (so breaking changes are in minor releases)

JLHwung: Yes, 0.x -> 0.x+1 is what I mean for a new major.

packages/babel-helper-define-polyfill-provider/src/index.ts
```diff
@@ -321,11 +321,29 @@ export default function definePolyfillProvider<Options>(
...
 
-      pre() {
+      pre(file) {
+        if (runtimeName) {
+          if (
+            file.get("runtimeHelpersModuleName") &&
+            file.get("runtimeHelpersModuleName") !== runtimeName
+          ) {
+            console.warn(
+              `Two different polyfill providers are trying to define two` +
```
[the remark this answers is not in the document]
nicolo-ribaudo: Done!

nicolo-ribaudo:
@JLHwung @liuxingbaoyu Any opinion regarding this?

PR #161, opened 2023-04-19
Update `core-js`

zloirock:


packages/babel-plugin-polyfill-corejs3/src/shipped-proposals.ts
```diff
@@ -4,6 +4,13 @@ export default new Set<string>([
   "esnext.array.from-async",
   "esnext.array.group",
   "esnext.array.group-to-map",
+  "esnext.array-buffer.detached",
+  "esnext.array-buffer.transfer",
+  "esnext.array-buffer.transfer-to-fixed-length",
+  "esnext.json.is-raw-json",
+  "esnext.json.parse",
+  "esnext.json.raw-json",
+  "esnext.set.difference.v2",
```
zloirock, line 13: Note that this module can be removed by `build` script, but should be here.

JLHwung: Can we fix the `build` script? So we don't have to remember that this module is a false negative.

JLHwung (approved):

zloirock:
@nicolo-ribaudo

PR #163, opened 2023-05-26
Fix corejs2 babel/runtime compat mode

nicolo-ribaudo:
cc @liuxingbaoyu 

https://github.com/babel/babel/pull/15660#discussion_r1207142190


liuxingbaoyu (approved):
Amazing! Thanks for the quick fix!

PR #165, opened 2023-06-15
Update `core-js`

zloirock:


packages/babel-plugin-polyfill-corejs3/test/fixtures/usage-pure/all-proposals-chrome-90-corejs-3.31/output.mjs
```diff
@@ -0,0 +1,137 @@
...
+
+_Symbol$isRegisteredSymbol(foo);
+
+_Symbol$metadata;
+
+_Object$groupBy(a, b);
+
+Map.groupBy(a, b);
+Promise.withResolvers();
+new URLSearchParams().has(a, b);
```
zloirock, line 137: Note that 3 lines above should be transpiled, but it's ignored with the current `usage-pure` optimization logic.

For example, `core-js-pure/features/promise` contain `esnext.promise.with-resolvers` that's missed everywhere - so `core-js-pure/features/promise` should be injected and replace `Promise`.

It's not directly related to this PR, it can be merged without it, I just show the problem.

nicolo-ribaudo:
Thank you!

PR #168, opened 2023-07-04
Use forked `semver@6` with backported security fixes

Jir4:
Hi,

For consistency purpose (and to get rid of warnings, I should admit), I think this project must use the newly backport forked of `node-semver`. I shamely just adapted the fix made by @nicolo-ribaudo on the main repo.

### DIFF
--- a/lib/semver.d.ts
+++ b/lib/semver.d.ts
@@ -1,0 +1,4 @@
+declare module "@nicolo-ribaudo/semver-v6" {
+  export { default } from "./semver";
+  export * from "./semver";
+}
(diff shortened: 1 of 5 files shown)


nicolo-ribaudo (dismissed):
Thank you :)

lib/semver.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "@nicolo-ribaudo/semver-v6" {
+  export { default } from "./semver";
+  export * from "./semver";
```
nicolo-ribaudo, line 3: This is a "circular dependency", it should re-export from the real `semver` package. If TS gives an error about `semver` not being found, you can add semver@7 as a devDependency of the top-level package.json

nicolo-ribaudo:
For the CI failure, please run `yarn dedupe`.

lib/semver.d.ts
Jir4: The change doesn't seem to bother `tscheck`

PR #181, opened 2023-09-18
Fix JSON.stringify polyfill name

### LINKED ISSUE #180: Excluding `es.json.stringify` does not work
Config:

```json
{
  "plugins": [
    ["polyfill-corejs3", { "method": "usage-pure", "exclude": ["es.json.stringify"] }]
  ]
}
```

Input:

```js
JSON.stringify({});
```

Expected output:

```js
JSON.stringify({});
```

Actual output:

```js
import _JSON$stringify from "core-js-pure/stable/json/stringify.js";
_JSON$stringify({});
```

mohd-akram:
Fixes #180


JLHwung (approved):
Thanks!

nicolo-ribaudo (approved):
Thank you!

mohd-akram:
Fixed the tests.

packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.ts
```diff
@@ -385,7 +385,7 @@ export const StaticProperties: ObjectMap2<CoreJSPolyfillDescriptor> = {
       "es.object.create",
       "es.object.freeze",
     ]),
-    stringify: define("json/stringify", ["es.json.stringify"], "es.symbol"),
+    stringify: define("json/stringify", ["es.json.stringify"]),
```
nicolo-ribaudo, line 388: It's possible that the reason there was `es.symbol` is that when polyfilling `JSON.stringify` we also have to polyfill symbols. In that case, this definition should be
```suggestion
    stringify: define("json/stringify", ["es.json.stringify", "es.symbol"]),
```

Let's wait for @zloirock to see this PR before merging :)

zloirock, line 388: It's not a bug or something like that. In early versions of `core-js@3`, `JSON.stringify` logic was only in `es.symbol` module to fix `JSON.stringify` with symbols. After that, many other fixes of  `JSON.stringify` logic, [well-formed unicode](https://github.com/tc39/proposal-well-formed-stringify), [`JSON.rawJSON`](https://github.com/tc39/proposal-json-parse-with-source) support, etc. were added and the main part of `JSON.stringify` polyfill was moved to `es.json.stringify`.

For modern `core-js` versions, `es.symbol` is not required, however, it could break compatibility with early `core-js` versions.

mohd-akram: Thanks, fixed.

packages/babel-plugin-polyfill-corejs3/test/fixtures/usage-pure/built-in-static-methods-proposals/output.js
```diff
@@ -2,7 +2,6 @@ var _Array$from = require("core-js-pure/features/array/from.js");
 var _Array$isArray = require("core-js-pure/features/array/is-array.js");
 var _Array$of = require("core-js-pure/features/array/of.js");
 var _Date$now = require("core-js-pure/features/date/now.js");
-var _JSON$stringify = require("core-js-pure/features/json/stringify.js");
```
zloirock, line 5: It should not be removed with [those](https://github.com/babel/babel-polyfills/blob/main/packages/babel-plugin-polyfill-corejs3/test/fixtures/usage-pure/built-in-static-methods-proposals/options.json) options.

mohd-akram: I think it's removed because it seems to default to version `3.0`. I noticed this while testing the package.

zloirock: [This entry is available in `3.0`](https://unpkg.com/core-js-pure@3.0.1/features/json/stringify.js).

IIRC it could be removed because of the order of modules in the array of dependencies - here the first is `es.json.stringify` that wasn't in `3.0` - however, it does not mean that `core-js-pure/features/json/stringify` is not required.

zloirock:
I think that the way solving of this issue is wrong - it's better to allow passing entries for exclusion, like `exclude: ['json/stringify']` or something like that.

packages/babel-plugin-polyfill-corejs3/test/fixtures/usage-pure/built-in-static-methods-proposals/output.js
mohd-akram: It checks `getModulesByVersion` (something like that, which references a JSON file with that metadata) in `core-js-compat` and `es.json.stringify` is not mentioned under `3.0`. I think that's why.

zloirock: Yes, because `JSON.stringify` polyfill was only in `es.symbol` and it was exported in `core-js-pure/features/json/stringify`.

mohd-akram:
This PR is to fix the exclusion :D I think perhaps the library should default to using the latest version of core-js rather than 3.0.

mohd-akram:
Any update on this?

zloirock:
Note that passing entries to include / exclude in `usage-pure` mode is allowed in `@core-js/babel-plugin` for `core-js@4`.

Issue #192, opened 2023-12-13
Sorting `core-js` modules in the order from `core-js-compat`


zloirock:
[As I wrote many times](https://github.com/babel/babel-polyfills/issues/80#issuecomment-808173774), removing the sorting logic was a mistake since some modules should be defined after dependencies.

At work on #191, I see that it could break typed array polyfills in old engines:

![image](https://github.com/babel/babel-polyfills/assets/2213682/492ba9ba-ae20-4ebe-9f44-eca6d2dffb50)

`es.typed-array.int8-array` should be loaded before any other typed array modules, so it was broken before #191 and will be broken after.

PR #194, opened 2024-01-15
Support injecting polyfill imports in a specific order

nicolo-ribaudo:
Fixes https://github.com/babel/babel-polyfills/issues/192

@zloirock Sorry that fixing this took so long. I have a question: do we also need to ensure ordering for `core-js-pure` polyfills, or only for `core-js`?


zloirock:
For `core-js-pure` it's not critical since in this case we work with entries and anyway, we haven't a data source for that -)

Issue #166, opened 2023-06-15
`usage-pure` has incorrect optimization logic


zloirock:
The problem is described [here](https://github.com/babel/babel-polyfills/pull/165#discussion_r1231253950).

zloirock:
https://github.com/babel/babel-polyfills/pull/198 fixes only a case for static methods. However, here also a case of prototype methods, by the link:
```js
new URLSearchParams().has(a, b);
```
should be transpiled for support 2 arguments in `.has` method.

liuxingbaoyu:
https://github.com/babel/babel-polyfills/pull/198/files#diff-ff6ba67b7a3296fe7754abd7eaf8d196a3b990236a03acd9e6b2bff37e5743e6R82

Just transpile `URLSearchParams` not working?

zloirock:
@liuxingbaoyu https://github.com/babel/babel-polyfills/pull/198/files#diff-cc59d35088698c787c120a9cdca94f7c0bde0d82bf01c79dd67e264226f0e8ffR98

PR #198, opened 2024-01-22
Add missing things in core-js-pure 3

liuxingbaoyu:
Fixes: #166

It seems that issues is not the optimization logic, but that we don't define the data of `corejs-pure`.


packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.ts
```diff
@@ -355,7 +355,7 @@ export const StaticProperties: ObjectMap2<CoreJSPolyfillDescriptor> = {
   },
 
   ArrayBuffer: {
-    isView: define(null, ["es.array-buffer.is-view"]),
+    isView: define("array-buffer/is-view", ["es.array-buffer.is-view"]),
```
zloirock, line 358: It's missed in the pure version.

packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.ts
```diff
@@ -434,10 +434,10 @@ export const StaticProperties: ObjectMap2<CoreJSPolyfillDescriptor> = {
   },
 
   Map: {
-    from: define(null, ["esnext.map.from", ...MapDependencies]),
-    groupBy: define(null, ["es.map.group-by", ...MapDependencies]),
-    keyBy: define(null, ["esnext.map.key-by", ...MapDependencies]),
-    of: define(null, ["esnext.map.of", ...MapDependencies]),
+    from: define("map/from", ["esnext.map.from", ...MapDependencies]),
```
zloirock, line 437: `.from` and `.of` are present in all collections.

zloirock:
Yes, it looks like a partial workaround.

packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.ts
```diff
@@ -515,17 +515,17 @@ export const StaticProperties: ObjectMap2<CoreJSPolyfillDescriptor> = {
...
-    any: define(null, [
+    any: define("promise/any", [
       "es.promise.any",
       "es.aggregate-error",
       ...PromiseDependenciesWithIterators,
     ]),
     race: define(null, PromiseDependenciesWithIterators),
-    try: define(null, ["esnext.promise.try", ...PromiseDependencies]),
+    try: define("promise/try", ["esnext.promise.try", ...PromiseDependencies]),
     withResolvers: define(null, [
```
zloirock, line 529: It's also a static method - it was added recently, but with `Map.groupBy` situation the same.

packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.ts
liuxingbaoyu: image
Is it unavailable?

packages/babel-plugin-polyfill-corejs3/test/fixtures/usage-pure/built-in-static-methods/output.js
```diff
@@ -193,3 +200,17 @@ Date.something;
...
+_Promise.all;
+_Promise$allSettled;
+_Promise$any;
+_Promise.race;
+_Promise.try;
+_Promise.withResolvers;
+_WeakMap.from;
+_ArrayBuffer$isView;
+Uint8Array.from;
+Uint8Array.of;
```
liuxingbaoyu, line 216: Is this expected? I didn't find the corresponding file.

packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.ts
zloirock: It's a cap.

https://github.com/zloirock/core-js/blob/master/packages/core-js-pure/override/modules/es.array-buffer.is-view.js

packages/babel-plugin-polyfill-corejs3/test/fixtures/usage-pure/built-in-static-methods/output.js
```diff
@@ -193,3 +199,17 @@ Date.something;
...
+_Map.from;
+_Map$groupBy;
+_Map.keyBy;
+_Map.of;
+_Promise.all;
+_Promise$allSettled;
+_Promise$any;
+_Promise.race;
+_Promise.try;
+_Promise.withResolvers;
```
liuxingbaoyu, line 211: This seems a bit strange.

nicolo-ribaudo (approved):

nicolo-ribaudo:
Thank you!

zloirock:
> `.from` and `.of` are present in all collections.

That means that those entries should be defined for all collections, not removed from `Map` -)

liuxingbaoyu:
> > `.from` and `.of` are present in all collections.
> 
> That means that those entries should be defined for all collections, not removed from `Map` -)

Oops, I just tested that `Map.from/Map.of` does not exist.

PR #199, opened 2024-01-24
Improve output when applying polyfills

liuxingbaoyu:
This assumes that the polyfill will not return false.
I'm not sure if this PR is safe, but it should be useful.


liuxingbaoyu:
Maybe we could add a `process.babel_polyfills_option_remove_dead_branch` to use it in the Babel repo first.

PR #201, opened 2024-01-25
[corejs3] Do no injected constructor polyfills for static props

### LINKED ISSUE #80: `Symbol.iterator` should not load full `Symbol` polyfill
```js
// Now:
arr[Symbol.iterator]();
// =>
import "core-js/modules/es.symbol.iterator.js";
import "core-js/modules/es.array.iterator.js";
import "core-js/modules/es.object.to-string.js";
import "core-js/modules/es.string.iterator.js";
import "core-js/modules/web.dom-collections.iterator.js";
import "core-js/modules/es.symbol.js";
import "core-js/modules/es.symbol.description.js";
arr[Symbol.iterator]();
```
`es.symbol` and `es.symbol.description` should not be loaded only on `Symbol.iterator` - it's too expensive. `core-js` architecture should allow to use `es.symbol.iterator` and iterators without those modules.

nicolo-ribaudo:
The first commit implements the capability, the second one uses it for core-js.

Fixes #80


packages/babel-plugin-polyfill-corejs3/test/fixtures/usage-global/typed-array-static/output.mjs
```diff
@@ -1,7 +1,6 @@
 import "core-js/modules/es.array.iterator.js";
 import "core-js/modules/es.array-buffer.slice.js";
 import "core-js/modules/es.object.to-string.js";
-import "core-js/modules/es.typed-array.int8-array.js";
```
nicolo-ribaudo, line 4: @zloirock Does `Int8Array.of();` need this polyfill?

zloirock: Yes, since in some cases required fixing or complete replacement of typed arrays.

liuxingbaoyu (approved):

liuxingbaoyu:
If we take this into consideration it might be better to implement "skip sub-paths".
But I thought that would be more complicated, and now it's fine to merge it.

packages/babel-plugin-polyfill-corejs3/test/fixtures/usage-global/typed-array-static/output.mjs
nicolo-ribaudo: I updated the deps list to add it back.

PR #208, opened 2024-03-20
Update `core-js`

zloirock:


scripts/corejs3-shipped-proposals-special-cases.mjs
```diff
@@ -1,4 +1,5 @@
 export default [
+  "esnext.suppressed-error.constructor",
```
zloirock, line 2: Because of some fixes for Bun implementation.

nicolo-ribaudo (approved):
Thanks!

PR #223, opened 2024-10-31
Update `core-js`

zloirock:


packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.ts
```diff
@@ -782,43 +789,39 @@ export const InstanceProperties = {
...
+    "es.iterator.drop",
+    ...IteratorDependencies,
     "esnext.async-iterator.drop",
     ...AsyncIteratorDependencies,
-    "esnext.iterator.drop",
-    ...IteratorDependencies,
-  ]),
-  emplace: define("instance/emplace", [
-    "esnext.map.emplace",
-    "esnext.weak-map.emplace",
```
zloirock, line 792: With the current approach, it should be injected on constructors call, and that's already happening. Here are missed dependencies and in this way anyway it could cause a error. However, in case of pure mode + targets, it could fail because of https://github.com/babel/babel-polyfills/issues/166.

packages/babel-plugin-polyfill-corejs3/src/built-in-definitions.ts
```diff
@@ -782,43 +789,39 @@ export const InstanceProperties = {
...
   ]),
   endsWith: define("instance/ends-with", ["es.string.ends-with"]),
   entries: define("instance/entries", ArrayNatureIteratorsWithTag),
   every: define("instance/every", [
     "es.array.every",
-    "esnext.async-iterator.every",
```
zloirock, line 798: Because lack of dependencies, it could cause a error, so this and similar cases commented. They are covered by `AsyncIteratorProblemMethods`.

JLHwung (approved):

PR #225, opened 2025-03-01
fix: Use `resolve` instead of `evaluate`

liuxingbaoyu:
Currently relies on buggy behavior.
https://github.com/babel/babel/pull/17156


nicolo-ribaudo (approved):
Btw, could you change the comment in the Babel repo for `resolve`? The current one (`Resolve a "pointer" `NodePath` to it's absolute path.`) doesn't make much sense.

JLHwung (approved):

packages/babel-helper-define-polyfill-provider/src/utils.ts
```diff
@@ -25,11 +21,9 @@ function resolveId(path): string {
...
   }
 
-  if (path.isPure()) {
-    const { deopt } = path.evaluate();
-    if (deopt && deopt.isIdentifier()) {
-      return deopt.node.name;
-    }
+  const resolved = path.resolve();
+  if (resolved.isIdentifier()) {
+    return resolved.node.name;
```
liuxingbaoyu, line 26: While working on https://github.com/babel/babel-polyfills/pull/225#pullrequestreview-2652557253 I realized there is a observable difference.
```
var b = Math;
b.f16round();
```
Now it won't return `Math`, maybe I should open a PR to restore this behavior?

nicolo-ribaudo: Yes, please!

PR #227, opened 2025-03-11
fix: Better resolve behavior

liuxingbaoyu:
Ref: https://github.com/babel/babel-polyfills/pull/225#discussion_r1986131558


packages/babel-plugin-polyfill-corejs3/test/fixtures/usage-pure/indirect-binding-2/output.mjs
```diff
@@ -0,0 +1,2 @@
+import Math from './math';
+Math.f16round(foo);
```
liuxingbaoyu, line 2: This is a new fix compared to `path.evaluate()`!

PR #239, opened 2025-07-01
Replace lodash with es-toolkit

dayongkr:
### Summary
This PR replaces `lodash.debounce` dependency in `babel-helper-define-polyfill-provider` and `lodash` in build scripts with `es-toolkit`, achieving significant bundle size reduction while maintaining full compatibility.

### Changes
- **babel-helper-define-polyfill-provider**: Replaced `lodash.debounce` with `es-toolkit/debounce`
- **Build scripts**: Replaced `lodash` with `es-toolkit`

### Benefits
- **Bundle size reduction**: debounce function code reduced from 378 lines to 110 lines (~3x smaller)
- **Cleaner implementation**: es-toolkit removes unnecessary utilities like `toNumber`, `isSymbol`, etc., keeping only the essential logic needed for most use cases
- **Zero breaking changes**: Build output diff shows identical results except for the migration itself
- **Full compatibility**: es-toolkit/compat passes all lodash test cases and maintains 100% interface compatibility

### Performance Impact
While debounce functions don't typically have significant performance implications, the substantial bundle size reduction (3x smaller) provides clear value for end users.

### Testing
- ✅ All existing tests pass
- ✅ Build output verification shows identical results

  - Screenshot 2025-07-02 at 12 07 55 AM

- ✅ es-toolkit/compat compatibility confirmed through lodash test suite

### Migration Details
The migration maintains the exact same API surface, ensuring this is a drop-in replacement with no changes required for consumers of this package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "gulp-watch": "^5.0.1",
     "jest": "^27.5.1",
     "lerna": "^3.22.1",
-    "lodash": "^4.17.21",
     "prettier": "^3.6.2",
     "rimraf": "^2.7.1",
     "rollup": "^2.79.2",
@@ -93,5 +92,8 @@
     }
   },
   "version": "0.0.0",
-  "packageManager": "yarn@4.9.2"
+  "packageManager": "yarn@4.9.2",
+  "dependencies": {
+    "es-toolkit": "^1.39.5"
+  }
 }
--- a/packages/babel-helper-define-polyfill-provider/package.json
+++ b/packages/babel-helper-define-polyfill-provider/package.json
@@ -36,7 +36,7 @@
     "@babel/helper-compilation-targets": "^7.27.2",
     "@babel/helper-plugin-utils": "^7.27.1",
     "debug": "^4.4.1",
-    "lodash.debounce": "^4.0.8",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -93,5 +92,8 @@
     }
   },
   "version": "0.0.0",
-  "packageManager": "yarn@4.9.2"
+  "packageManager": "yarn@4.9.2",
+  "dependencies": {
```
nicolo-ribaudo, line 96: This should be a devDependency

dayongkr: Thanks for catching this!

dayongkr:
> Also sorry, but the `es-toolkit` npm package is 35x larger than the `lodash.debounce` (`lodash` is just a devDependency, so it doesn't really count):
> 
> * https://bundlephobia.com/package/lodash.debounce@4.0.8
> * https://bundlephobia.com/package/es-toolkit@1.39.5

While it's true that downloads would increase, since you can import only the functions(files) you need, the actual build output will be smaller than before.

dayongkr:
For now, es-toolkit targets Node LTS versions, but I haven't measured the lodash compatibility layer like es-toolkit/compat.

I'll look this part, align it with simillar specs as lodash, and share the results through this PR again!

PR #248, opened 2026-03-05
[deps] update `resolve`

ljharb:
This updates to resolve@next, which for this package's uses is a nonbreaking update.

What this allows in the future, however, is `exports`-aware resolution.

### DIFF
--- a/packages/babel-helper-define-polyfill-provider/package.json
+++ b/packages/babel-helper-define-polyfill-provider/package.json
@@ -37,7 +37,7 @@
     "@babel/helper-plugin-utils": "^7.28.6",
     "debug": "^4.4.3",
     "lodash.debounce": "^4.0.8",
-    "resolve": "^1.22.11"
+    "resolve": "^2.0.0-next.6"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


JLHwung:
@ljharb Could you rebase and run `yarn dedupe`?

ljharb:
hm, this package no longer seems to include `resolve`. What does it use for resolution now?

JLHwung:
> hm, this package no longer seems to include `resolve`. What does it use for resolution now?

Previously `resolve` serves as polyfill for Node.js < 8.9 which does not support the `paths` option. Now that the library drops Node.js < 20 support, it now only uses the native `require.resolve`.

ljharb:
do you need to resolve anything based on a range of user-supported node versions? (even if the tool requires node >= 20)

JLHwung:
> do you need to resolve anything based on a range of user-supported node versions? (even if the tool requires node >= 20)

I don't recall we have such an API to resolve something in a way that Node.js 6 would resolve.

@nicolo-ribaudo Do you think there is a use case for that?

ljharb:
If so, you'll want resolve@next, if not, this PR should be closed :-)

